PR #6, opened 2019-11-18
Add selenium and card smoke-test pipelines

rjbaker:
Adds:

* `selenium-hub` pipeline - Deploys a persistent Selenium hub / chrome application to PaaS `tools` space 
* `smoke-tests` pipeline - Runs card smoke tests on Concourse using existing pay-endtoend docker container using selenium hub running on PaaS

### DIFF
--- a/ci/pipelines/smoke-tests.yml
+++ b/ci/pipelines/smoke-tests.yml
@@ -1,0 +1,34 @@
+---
+groups:
+  - name: Smoke Tests
+    jobs:
+      - card-smoke-test
+
+resources:
+  - name: every-10m
+    type: time
+    source: { interval: 10m }
+
+  - name: omnibus
+    type: git
+    source:
+      uri: https://github.com/alphagov/pay-omnibus
+      branch: smoke_tests
+
+jobs:
+  - name: card-smoke-test
+    serial: true
+    plan:
+      - get: every-10m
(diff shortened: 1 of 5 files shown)


ci/pipelines/smoke-tests.yml
```diff
@@ -0,0 +1,34 @@
...
+  - name: card-smoke-test
+    serial: true
+    plan:
+      - get: every-10m
+        trigger: true
+      - get: omnibus
+      - task: test
+        file: omnibus/ci/tasks/run-smoke-test.yml
+        params:
+          SMOKE_TEST: smoke-card
```
danworth: Do we need to set the `PUBLICAPI_URL` and `SELFSERVICE_URL` vars here too? I'm looking at the params for `ci/tasks/run-smoke-test.yml` that seem to require setting:
```
params:
  SMOKE_TEST:
  PUBLICAPI_URL:
  SELFSERVICE_URL:
  SELENIUM_HUB_URL:
```

rjbaker: I'm not 100% sure what params each smoke test needs just yet. These are just placeholders for now.

danworth: oh right sorry, makes sense 👍

PR #17, opened 2019-11-27
PP-5876: add notifications service to paas skeleton

rjbaker:
Adds the notifications/naxsi service in front of connector & dd-connector.
The notifications container required some modifications to run on PaaS. 

Specifically:
* Disable default nginx ipv6 listener as this isn't supported and causes the container to crash
* Bind nginx DNS resolver to PaaS DNS host. This is now configurable.
* Removed self-signed SSL cert config. This isn't needed as envoy should handle this implicitly.
* Configure Tini (init script) to run as a subreaper since CF already has an init system.

Notifications container changes are on the following branch: https://github.com/alphagov/pay-notifications/tree/paas_configuration

Paired with @nimalank7

### DIFF
--- a/paas/pay-apps.yml
+++ b/paas/pay-apps.yml
@@ -329,8 +329,8 @@ applications:
       <<: *dockerImage
       image: govukpaypaas/notifications:latest-master
     env:
-      CONNECTOR_URL: https://card-connector-((space)).apps.internal
-      DD_CONNECTOR_URL: https://directdebit-connector-((space)).apps.internal
+      CONNECTOR_URL: http://card-connector-((space)).apps.internal:8080
+      DD_CONNECTOR_URL: http://directdebit-connector-((space)).apps.internal:8080
       TINI_SUBREAPER: ''
     routes:
     - route: pay-notifications-((space)).london.cloudapps.digital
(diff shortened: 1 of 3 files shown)


paas/pay-apps.yml
```diff
@@ -322,6 +323,17 @@ applications:
...
     - route: pay-selfservice-((space)).london.cloudapps.digital
+  - <<: *applicationTemplate
+    name: notifications
+    docker:
+      <<: *dockerImage
+      image: govukpaypaas/notifications:latest-master
+    env:
+      CONNECTOR_URL: http://card-connector-((space)).apps.internal:8080
+      DD_CONNECTOR_URL: http://directdebit-connector-((space)).apps.internal:8080
+      TINI_SUBREAPER: ''
```
stephencdaly: what does this environment variable do?

vixus0: When you run a Docker container in PaaS it's actually being run by a tool called Garden which takes over as the PID1 process inside the container. Any zombie processes (subprocesses where the parent process has ended) in the container get reparented to the PID1 process. 

Setting `TINI_SUBREAPER` means it's able to have zombie processes reparented to it even if it's not PID1. I'm assuming it only gets zombie processes that are themselves a subsubsubprocess of `tini`.

PR #28, opened 2019-12-09
PP-5907: Create separate staging deploy pipeline

rjbaker:
This change adds a new `deploy.yml` pipeline. This is used to deploy built artefacts for each application to the Staging PaaS environment. This also decouples the staging environment CF space definition from the dev environment spaces.

At a later point this pipeline may be updated to further promote built artefacts to a production environment following successful deployment validation (smoke tests etc).

Currently, application deploys are triggered by changes to an application Docker image tagged with `latest-master`. This may later be changed to respond to git tags promoting a build from a successful CI test stage.

Screenshot 2019-12-10 at 13 51 59

### DIFF
--- a/ci/pipelines/deploy.yml
+++ b/ci/pipelines/deploy.yml
@@ -1,0 +1,270 @@
+---
+resource_types:
+  - name: cf-cli
+    type: docker-image
+    source:
+      repository: nulldriver/cf-cli-resource
+
+resources:
+  - name: omnibus
+    type: git
+    icon: github-circle
+    source:
+      uri: https://github.com/alphagov/pay-omnibus
+      branch: develop
+
+  - name: paas-staging
+    type: cf-cli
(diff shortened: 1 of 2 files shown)


ci/pipelines/deploy.yml
```diff
@@ -0,0 +1,416 @@
...
+
+  - <<: *image
+    name: notifications
+    source:
+      <<: *image-source
+      repository: govukpaypaas/notifications
+
+jobs:
+  - name: deploy-adminusers-staging
+    serial_groups: [adminusers-stg]
```
stephencdaly: Should all deploy jobs be part of the same serial group, as we probably want only one app to be deployed at a time? Or is that handled in a different way?

rjbaker: Probably one for the team to discuss as there are cases where various apps can safely be deployed together.

PR #33, opened 2020-01-07
PP-5906: Staging PaaS environment

vixus0:
This is the work to create a staging environment with a separate CDE space for the apps
that deal with cardholder data.

- Create script for provisioning PaaS environments `paas/provision_space.sh`
- Point staging Cloudfront to Ireland PaaS
- Manually created new staging PaaS environment
- Fix routes in PaaS manifest
- Change `deploy` pipeline to push apps to new staging environment
- Refactor Concourse pipelines to use script for provisioning dev envs
- Deploy smoke test apps to Ireland


mrlumbu:
Looks good. My personal preference would be to have configuration in yaml or similar rather than directly in bash - I can foresee that bash script spidering out into something unmanageable.

PR #43, opened 2020-01-23
Document Terraform setup and usage

rjbaker:
Adds a basic guide to setting up the project Terraform from scratch.

### DIFF
--- a/docs/source/infrastructure.html.md.erb
+++ b/docs/source/infrastructure.html.md.erb
@@ -1,0 +1,50 @@
+---
+title: Infrastructure
+---
+
+# Infrastructure
+
+Pay uses Terraform to manage all infrastructure residing in AWS staging & production environments.
+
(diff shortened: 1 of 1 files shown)


docs/source/infrastructure.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+
+```sh
+mv terraform-provider-pass ~/.terraform.d/plugins
+```
+
+## Run Terraform with AWS Credentials
+
+Ensure you have access to the Pay AWS account by assuming a role delegated via the gds-users account.
+
+Use [aws-vault](https://github.com/99designs/aws-vault) to manage credentials and execute commands using temporary credentials.
```
richardTowers: Current RE recommendation is actually to use `gds-cli` for this (which uses aws-vault under the hood, so no real difference). Could link out to https://reliability-engineering.cloudapps.digital/iaas.html#use-the-command-line

I think recommending aws-vault is fine until we've set that up though.

PR #44, opened 2020-01-24
PP-6030 Switch card-frontend to deploy from buildpack

danworth:
Add a github-release resource to watch for releases with appropriate
tag. Once triggered the `download_deployment_artifact.yml` task will get the
zipped archive and extract it so that the `cf` resource can push it
using the included manifest.

This commit will require a new credential on concourse for the github
api token, referred to within the deploy job as `git-access-token`.

**What**
I've tested something very similar on a local pipeline pushing to my dev space so should work. Just doing it for Frontend to begin with.

~This is dependent upon Richard's work and adding the new github access token cred to concourse, hence the `do-not-merge`.
https://github.com/alphagov/pay-omnibus/pull/45#pullrequestreview-347990689~ Richard's pr is merged.

### DIFF
--- a/ci/tasks/download-deployment-artifact.yml
+++ b/ci/tasks/download-deployment-artifact.yml
@@ -1,0 +1,20 @@
+platform: linux
+
+image_resource:
+  type: registry-image
+  source:
+    repository: alpine
+
+inputs:
+  - name: git-release
+    optional: false
+
+outputs:
+  - name: artifact
+
+run:
+  path: sh
+  args:
+    - -c
+    - |
+      tar -C artifact -xf git-release/pay-*-*.tar.gz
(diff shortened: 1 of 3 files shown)


ci/tasks/download-deployment-artifact.yml
```diff
@@ -0,0 +1,20 @@
+platform: linux
+
+image_resource:
+  type: registry-image
+  source:
+    repository: alpine
+
+inputs:
+  - name: git-release
+    optional: false
```
mrlumbu: is this not `false` by default?

danworth: It didn't say in the docs what the default is and then I though this make its very clear to anyone looking at it so I kept it. If you're right then I'm happy to remove it?
https://concourse-ci.org/tasks.html#input-optional

mrlumbu: yeah, I actually looked at docs and it isn't clear. I just think it is but happy to leave in since it is clearer

PR #48, opened 2020-01-27
PP-6030 Switch node apps to buildpack deployment

danworth:
- selfservice, directdebit-frontend and products-ui

### DIFF
--- a/ci/pipelines/deploy.yml
+++ b/ci/pipelines/deploy.yml
@@ -73,7 +73,7 @@ resources:
 
   - name: git-card-frontend-pre-release
     type: github-release
-    source:
+    source: &github-release-source
       owner: alphagov
       access-token: ((github-access-token))
       tag_filter: "paas_release-(.*)"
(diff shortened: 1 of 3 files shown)


ci/pipelines/deploy.yml
```diff
@@ -304,15 +304,22 @@ jobs:
     serial_groups: [directdebit-frontend-stg]
     plan:
       - get: omnibus
-      - get: directdebit-frontend
+      - get: git-directdebit-frontend-pre-release
         trigger: true
+      - task: download-directdebit-frontend-artifact
+        file: omnibus/ci/tasks/download-deployment-artifact.yml
```
vixus0: I thought the github-release resource would download artifacts for you:
https://github.com/concourse/github-release-resource#in-fetch-assets-from-a-release

ci/pipelines/deploy.yml
```diff
@@ -304,15 +304,22 @@ jobs:
...
+        input_mapping: { git-release: git-directdebit-frontend-pre-release }
+        output_mapping: { artifact: directdebit-frontend-artifact }
+      - put: paas-staging
         params:
```
vixus0: Maybe make this a YAML anchor and then you can use generic `path: artifact` and `manifest: artifact/manifest.yml` (don't think you even need the latter as it's the default).

danworth: Oh ok, I thought that the inputs and outputs were name spaced at the plan level so we'd need unique names for every archive per job (hence the mapping). Otherwise yes, using generic names will allow us to use an anchor and simplify it all. 

The `manifest` may actually be a hangover from the `cf` resource I was originally using where it's mandatory. I can try removing it when using the `cf-cli`.

ci/pipelines/deploy.yml
danworth: yeah it does and we're taking it from there, the name is misleading, it would be better named something like `extract-artefact` since that's what its doing.. naming things is hard... I'll correct it.
https://github.com/alphagov/pay-omnibus/blob/develop/ci/tasks/download-deployment-artifact.yml

PR #50, opened 2020-01-27
PP-6046 - Use Terraform to provision Pay PaaS environment

vixus0:
Terraform is a better option for provisioning the "infrastructure" in PaaS.
It also means we can properly automate provisioning the environment using
Concourse.

This makes use of the [Cloud Foundry provider for 
Terraform](https://github.com/cloudfoundry-community/terraform-provider-cf) 
which needs to be downloaded and installed in the Terraform plugins 
directory.

We've added the following new Terraform modules:

- `paas`: Defines the placeholder apps, network policies and the app
catalog service
- `paas-postgres`: Defines the RDS service instances
- `paas-stub-postgres`: Defines a stub postgres service for dev envs
- `paas-stub-sqs`: Defines a stub SQS service for dev envs

Deployments for the dev environments and the staging environment are in `dev-paas` and `staging-paas`, respectively.

For the `stub` modules, the `paas` module needs to be provisioned first:

    terraform init
    terraform apply -target=module.paas
    terraform apply


richardTowers:
> I'd like to also test the dev module before I approve it.

Tested the dev module this morning - comes up cleanly and it's very fast (since it doesn't have to do any databases). Very cool.

It will be interesting to see how long it takes to deploy the applications into this skeleton.

PR #67, opened 2020-01-30
PP-6068: Document how we will push cardid-data

richardTowers:


docs/source/adrs/0005-inline-cardid-data-into-cardid-before-pushing-to-gov-uk-paas.md
```diff
@@ -0,0 +1,62 @@
...
+artefact (a .jar file) to be in a public GitHub release. This means we cannot
+include the sensitive data in the build artefact.
+
+## Decision
+
+When concourse deploys cardid to GOV.UK PaaS it will:
+
+* download the cardid build artefact from github releases
+* clone the pay-cardid repository recursively (which will include the private pay-cardid-data)
+* insert the data from pay-cardid-data into the cardid build artefact
```
mrlumbu: just to be super clear - this means unzipping the build artefact and updating the jar with the relevant files

docs/source/adrs/0005-inline-cardid-data-into-cardid-before-pushing-to-gov-uk-paas.md
```diff
@@ -0,0 +1,62 @@
...
+The deployed package and droplet in GOV.UK PaaS will contain the pay-cardid-data.
+The package and droplet are created as part of
+[the cloudfoundry application staging process](https://docs.cloudfoundry.org/concepts/how-applications-are-staged.html),
+and stored in a private S3 bucket owned by GOV.UK PaaS. Only people with
+SpaceDeveloper access to the cloudfoundry space where the app is running and
+GOV.UK PaaS administrators can access packages and droplets.
+
+We will need to take care to ensure that the we use the commit of the
+pay-cardid-data submodule that matches the version in the pay-cardid GitHub
+release.
```
mrlumbu: Just thinking if this has any implications for managing paas dev environments. I guess `cf push`ing from laptop will be fine - just compile the jar locally with correct submodule checked out and push it. If for some reason you did want a particular deployed artefact (or as close as you can get), you would have to download the relevant release and mess around with it, then push it. But then I suppose for existing releases you could use concourse to do that for you.

docs/source/adrs/0005-inline-cardid-data-into-cardid-before-pushing-to-gov-uk-paas.md
```diff
@@ -0,0 +1,62 @@
...
+
+The [pay-cardid](https://github.com/alphagov/pay-cardid) microservice is
+responsible for retrieving card information for a given card number.
+
+To do this, it requires a number of proprietary .csv files, which are provided
+by various payment service providers. This data is sensitive in the sense
+that it's encumbered by licence restrictions, so we need to take appropriate care
+not to make it public.
+
+The data is roughly 36 MB uncompressed, and changes rarely (a few times a year).
```
mrlumbu: perhaps worth noting that it is a goal to make this much more frequent (or perhaps not relevant here)

docs/source/adrs/0005-inline-cardid-data-into-cardid-before-pushing-to-gov-uk-paas.md
vixus0: I don't think Maven adds the cardid data to the JAR, and the only thing you can push to CF is the JAR, so just doing `mvn package; cf push` won't work.

vixus0:
By chucking a manifest

```yaml
applications:
  - name: cardid-data
    buildpacks:
      - staticfile_buildpack
    routes:
      - route: cardid-data.apps.internal
```

into the cardid-data repo, it can be pushed as an internal app and then the data becomes accessible to permitted apps at something like `http://cardid-data.apps.internal:8080/test-cards/test-card-bin-ranges.csv`.

This could be a way to get around the need to manually put the data in the JAR and would avoid breaking `cf push` from the cardid repository.

docs/source/adrs/0005-inline-cardid-data-into-cardid-before-pushing-to-gov-uk-paas.md
rjbaker: As @mrlumbu pointed out the data will be dropped into the .jar, but I guess that means paths will need to be tweaked a little. Currently the Dockerfile build drops the data in a `/data` directory adjacent to the jar file. https://github.com/alphagov/pay-cardid/blob/master/Dockerfile#L22

docs/source/adrs/0005-inline-cardid-data-into-cardid-before-pushing-to-gov-uk-paas.md
richardTowers: Yes, that's right. `zip` and `tar` have options to append data to an archive without extracting it, but it's the same idea.

mrlumbu:
@vixus0 that's a pretty nice solution, definitely worth exploring. More or less treating bin range data as a big secret, which is an approach @richardTowers has mentioned before

richardTowers:
I've pushed a new commit which adds another ADR for @vixus0's suggestion (as I understand it), with my understanding of the consequences. We should discuss which approach we prefer and then amend the files so it's clear which one we're choosing.

PR #69, opened 2020-01-31
PP-6085 Add the env-map.yml to the release archive

danworth:
env-map.yml is used by the PaaS env-map-buildpack to flatten the VCAP services
vars for bound services on PaaS so that we don't have to change how our
apps consume env vars.


richardTowers:
Do we still need this now we've hacked the env-map.yml files into `src/main/resources` @danworth ?

danworth:
No longer needed.

PR #70, opened 2020-02-03
PP-6085 Add staging-cde creds

danworth:
It isn't possible to bind services to apps in another space so this
creates a version of the app-catalog for the staging-cde space. This
also adds `selfservice_transactions_url` which is needed by products-ui.

## WHAT
We deployed this on Friday running tf on my machine but didn't actually get round to committing it.

### DIFF
--- a/terraform/modules/paas/app-catalog-service.tf
+++ b/terraform/modules/paas/app-catalog-service.tf
@@ -1,31 +1,31 @@
 locals {
   app_catalog_urls = {
-    "adminusers_url"            = "http://${cloudfoundry_route.adminusers.endpoint}:8080"
-    "cardid_url"                = "http://${cloudfoundry_route.cardid.endpoint}:8080"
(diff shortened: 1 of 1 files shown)


terraform/modules/paas/app-catalog-service.tf
```diff
@@ -1,6 +1,29 @@
 resource "cloudfoundry_user_provided_service" "app_catalog" {
   name  = "app-catalog"
   space = data.cloudfoundry_space.space.id
+  credentials = {
```
rjbaker: This could be simplified a little by adding the credentials map to a local variable e.g.

```
locals {
  app_catalog_urls = {
    "adminusers_url" ...
  }
}
```

Then pass it into the `credentials = local.app_catalog_urls`

danworth: yeah that's a good idea...hold on I'll change it.

PR #73, opened 2020-02-05
PP-6085: Add SQS user proviced service inside the CDE space

richardTowers:
This is needed if we want connector to be able to talk to SQS in a nice way.

When applying the terraform we also noticed a weird edge case where terraform
wanted to turn all the health_check_type's back to "port" from "http" - we
think that might be a bug in the provider, but we're working around it for now
by telling terraform to ignore_changes.


rjbaker:
Looks good. Suggest renaming the module from `/paas-stub-sqs/` when SQS becomes a "thing".

PR #85, opened 2020-02-10
PP-6085: Make Publicauth service private

rjbaker:
### Context
Publicauth is in fact a **private** service. 
It does not require a public hostname or CloudFront distribution.

**This PR:**
- Removes public hostname & cloudfront distribution
- Adds internal route for publicauth
- Adds network policies to publicauth, selfservice & toolbox allowing connectivity to publicauth
- Adds further routes & urls to app-catalog user-provided services required by apps wishing to connect to publicauth

Apps will require restaging to update environment variables. 
env-map.yml may need to be updated in various apps to map additional routes from app-catalog to internal PUBLIC_AUTH_URL env var since this is inconsistently formatted across apps. These will be raised in separate PRs.


terraform/modules/paas/app-catalog-service.tf
```diff
@@ -19,6 +19,9 @@ locals {
     "products_ui_redirect_url"     = "https://${cloudfoundry_route.products_ui.endpoint}/redirect"
     "products_ui_pay_url"          = "https://${cloudfoundry_route.products_ui.endpoint}/pay"
     "products_ui_confirmation_url" = "https://${cloudfoundry_route.products_ui.endpoint}/payment-complete"
+    // Some apps require FQP. These are mapped in env-map.yml in respective app repos.
```
richardTowers: FQP = fully qualified path?

PR #92, opened 2020-02-11
Add external route to toolbox app

rjbaker:
Adds a PaaS-provided route to get access the toolbox app in the Staging environment. This will be accessible via the cloudapps.digital domain for now and provides https by default.

Todo:
- [ ] Add route service proxy to filter requests by whitelisted IP
- [ ] Add Cloudfront distro if we want to use a custom domain, this is required to support TLS

### DIFF
--- a/terraform/modules/paas/toolbox.tf
+++ b/terraform/modules/paas/toolbox.tf
@@ -9,8 +9,8 @@ resource "cloudfoundry_app" "toolbox" {
 }
 
 resource "cloudfoundry_route" "toolbox" {
-  domain   = data.cloudfoundry_domain.external.id
-  hostname = "toolbox${var.external_hostname_suffix}"
+  domain   = data.cloudfoundry_domain.paas_external.id
+  hostname = "toolbox${var.internal_hostname_suffix}"
   space    = data.cloudfoundry_space.space.id
 
   target {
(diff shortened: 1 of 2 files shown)


terraform/modules/paas/toolbox.tf
```diff
@@ -9,8 +9,8 @@ resource "cloudfoundry_app" "toolbox" {
 }
 
 resource "cloudfoundry_route" "toolbox" {
-  domain   = data.cloudfoundry_domain.external.id
-  hostname = "toolbox${var.external_hostname_suffix}"
+  domain   = data.cloudfoundry_domain.paas_external.id
+  hostname = "toolbox${var.internal_hostname_suffix}"
```
richardTowers: It looks kind of confusing having `paas_external` and the `internal_hostname_suffix` - I guess this is because the internal suffix is more "global", because it's a shared domain?

Probably fine for this PR, but I wonder if we want a better name for this variable (e.g. like `global_hostname_suffix`)?

rjbaker: I'd agree, although i'd rather not introduce another variable for this single url. Since we just need to namespace by the space, we could simply use `${var.space}?`

richardTowers: Nah, I think let's leave it with `internal_hostname_suffix`. Just wanted to check.

PR #93, opened 2020-02-12
PP-6127 Add secret service for publicapi

danworth:
Implement version 1 of provisioning secrets in paas. It is likely that
this will be replaced when PaaS provide a more robust means for secrets
management.

This adds a credentials module which will gather secrets from
pay-low-pass. The credentials are set on a user provided service which
will be bound to the app.

To get it started this commit deals with publicapi but the same approach
and be used for the other apps.

with @rjbaker

## WHAT
We intend to iterate and improve upon this but this is a start. Comments and feedback welcomed to help us improve this as we roll it out for the other apps.

### DIFF
--- a/terraform/modules/credentials/main.tf
+++ b/terraform/modules/credentials/main.tf
@@ -1,0 +1,17 @@
+locals {
+  /* This needs some careful checking. If keys() and values() do not 
+  return the corresponding hash keys and values in the same order then the
+  zipped result will have the wrong value for a key. Perhaps there's a better way*/
+  low_pass_values = zipmap(keys(data.pass_password.secret), values(data.pass_password.secret).*.password)
+}
+
+provider "pass" {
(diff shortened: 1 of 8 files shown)


terraform/modules/credentials/main.tf
```diff
@@ -0,0 +1,17 @@
+locals {
+  /* This needs some careful checking. If keys() and values() do not 
+  return the corresponding hash keys and values in the same order then the
+  zipped result will have the wrong value for a key. Perhaps there's a better way*/
+  low_pass_values = zipmap(keys(data.pass_password.secret), values(data.pass_password.secret).*.password)
```
rjbaker: I * think * this should work okay. The docs regarding `values`:

> The values are returned in lexicographical order by their corresponding keys, so the values will be returned in the same order as their keys would be returned from keys

PR #96, opened 2020-02-13
PP-6127 Add cardid-secret-service

danworth:


terraform/staging-paas/terraform.tfvars
```diff
@@ -41,4 +41,14 @@ credentials = {
       stripe_transaction_fee_percentage = "0.1"
     }
   }
+  cardid = {
+    pay_low_pass_secrets = {
+      sentry_dsn = "sentry/cardid_dsn"
+    }
+    static_values = {
+      test_card_data_location = "http://cardid-data-stg.apps.internal:8080/test-cards/test-card-bin-ranges.csv"
```
rjbaker: As discussed the host/port should probably be derived from data provided by the `app-catalog` service. The config will then only require the paths. We can leave as-is for now, but add card-id-data to app catalog now and refactor this later.

PR #106, opened 2020-02-13
PP-6127: Add directdebit connector secret service

rjbaker:
Adds user-provided CF service to provide creds to dd-connector app.

DD connector creds are sourced from both pay-low-pass and pay-dev-pass. This PR adds support for sourcing creds from both stores and merges the results into a single "secrets" output from the `credentials` module.

I have already applied the TF successfully in the staging env.


terraform/modules/credentials/outputs.tf
```diff
@@ -1,4 +1,4 @@
 output "secrets" {
   sensitive = true
   value = local.low_pass_values
```
danworth: Just checked in staging and the `directdebit-connector-secret-service` doesn't have the values from `pay-dev-pass`. I think it's because this output needs changing to merge in the `local.dev_pass_values` else they won't be available; this is only returning the `low_pass_values`

PR #108, opened 2020-02-14
PP-6127 Provision selfservice secrets service

mrlumbu:
Passing in secrets from pay-low-pass and pay-dev-pass

### DIFF
--- a/terraform/modules/paas/selfservice.tf
+++ b/terraform/modules/paas/selfservice.tf
@@ -1,3 +1,7 @@
+locals {
+  selfservice_credentials = lookup(var.credentials, "selfservice")
+}
+
 resource "cloudfoundry_app" "selfservice" {
   name    = "selfservice"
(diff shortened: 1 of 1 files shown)


terraform/modules/paas/selfservice.tf
```diff
@@ -50,3 +50,15 @@ resource "cloudfoundry_network_policy" "selfservice" {
     port            = "8080"
   }
 }
+
+module "selfservice_credentials" {
+  source = "../credentials"
+  pay_low_pass_secrets = lookup(local.selfservice_credentials, "pay_low_pass_secrets")
```
rjbaker: local.selfservice_credentials needs to be defined at the top. See: https://github.com/alphagov/pay-omnibus/blob/develop/terraform/modules/paas/directdebit-connector.tf#L1

mrlumbu: fixed

PR #122, opened 2020-03-03
Fix node_module cache behaviour when removing dev dependencies

rjbaker:
Caches ~/.npm global directory reducing the number of packages that are re-downloaded between dev and prod dependency installs. Props to @danworth 👍

### DIFF
--- a/ci/tasks/node-build.yml
+++ b/ci/tasks/node-build.yml
@@ -14,7 +14,8 @@ outputs:
   - name: build-output
 
 caches:
-  - path: src/node_modules
+  - path: node_module_cache_dev
+  - path: node_module_cache_prod
 
 params:
   LD_LIBRARY_PATH: /src/node_modules/appmetrics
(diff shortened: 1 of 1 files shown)


rjbaker:
> I think we could mount the cache into the container into some location other than `node_modules` (default is `~/.npm`) then update the npm config in the container to know where it is:
> https://docs.npmjs.com/misc/config#cache. Then when running npm install it'll pull from the cache.
> Then I think we've separated the app's `node_modules` directory from any of the caches/mounted directories and we maintain a single cache and don't have to use the symlinks etc.

Concourse cache paths in the task definition are only relative to the build environment. I don't think it's possible to target a path outside the build environment e.g. `~/.npm` in which case we need to start messing around with the package.json file to target a relative path.

richardTowers:
Could we do `npm config set cache some-relative-dir-that-is-cached` (or something like that?)

ci/tasks/node-build.yml
```diff
@@ -14,7 +14,7 @@ outputs:
   - name: build-output
 
 caches:
-  - path: src/node_modules
+  - path: ../../../root/.npm
```
danworth: Just wondering if this is susceptible to changing under our feet since it's outside the build environment (from what I understand). Maybe something to think about for the future.

ci/tasks/node-build.yml
```diff
@@ -33,7 +33,7 @@ run:
       npm rebuild node-sass
       npm install
       npm run compile
-      rm -rf ./node_modules
+      rm -rf node_modules
       npm install --only=prod
```
danworth: Shame that `--only=prod` doesn't remove any existing dev dependencies within node_modules (well the docs don't mention it does and I haven't tried it).

rjbaker: Yeah, this was the original problem. I assumed that's what it did, but apparently not and @richardTowers confirmed it.

PR #124, opened 2020-03-10
Add ADR on running e2e tests in Concourse

mrlumbu:


heathd:
This needs to reference the spike outputs, which tools we considered, what the pros and cons of each were.

I'm specifically interested in evidence that the docker-in-docker approach for concourse works and estimates of the sizing of machines or estimation of what size concourse cluster we need

mrlumbu:
@heathd not sure if an in depth consideration of pros and cons etc is appropriate for an adr http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions
Maybe we need an adr on adrs.

docs/source/adrs/0007-use-concourse-to-run-e2e-tests.md
```diff
@@ -0,0 +1,52 @@
...
+1. Using Concourse to run the tests against a target deployed on the PaaS <https://payments-platform.atlassian.net/browse/PP-6206>
+1. Using Concourse to run the tests against a target deployed on docker-compose
+1. Using Buildkite <https://payments-platform.atlassian.net/browse/PP-6214>
+
+For detailed discussion of the outcomes of the spikes see the relevant tickets.
+
+The conclusion from these spikes was that options 3 and 4 were both viable. 
+Option 3 is the most inline with GDS tech direction, but requires a bit more work to get the required performance - it is
+possible to run many builds in parralel on separate worker boxes, but this comes less natively than in option 4.
+Otion 4 is a good technological fit, but requires more work to get everything set up well, and is a distinct divergence from
```
rjbaker, line 41: Option typo

docs/source/adrs/0007-use-concourse-to-run-e2e-tests.md
```diff
@@ -0,0 +1,52 @@
...
+current GDS technology.
+
+## Decision
+We will run the e2e tests in docker compose (as they currently are), on Concourse (option 3 above).
+We will also move the rest of Pay's tests on to Concourse.
+We will arrange for a sufficient number of concourse workers to be proviioned for Pay to ensure we are not blocked by e2e tests.
+## Consequences
+
+- We will have to move the unit/integration tests currently on travis to concourse
```
rjbaker, line 50: We don't necessarily have to move these to Concourse, but it would make it easier to move build artefacts i.e. containers around for subsequent E2E tests. We also get access to the local ECR repo from Concourse.

docs/source/adrs/0007-use-concourse-to-run-e2e-tests.md
```diff
@@ -0,0 +1,52 @@
...
+For detailed discussion of the outcomes of the spikes see the relevant tickets.
+
+The conclusion from these spikes was that options 3 and 4 were both viable. 
+Option 3 is the most inline with GDS tech direction, but requires a bit more work to get the required performance - it is
+possible to run many builds in parralel on separate worker boxes, but this comes less natively than in option 4.
+Otion 4 is a good technological fit, but requires more work to get everything set up well, and is a distinct divergence from
+current GDS technology.
+
+## Decision
+We will run the e2e tests in docker compose (as they currently are), on Concourse (option 3 above).
```
heathd, line 45: this is using docker-in-docker? and we've proven that this works, right?

docs/source/adrs/0007-use-concourse-to-run-e2e-tests.md
```diff
@@ -0,0 +1,52 @@
...
+We will also move the rest of Pay's tests on to Concourse.
+We will arrange for a sufficient number of concourse workers to be proviioned for Pay to ensure we are not blocked by e2e tests.
```
heathd, line 47: do we know how concourse handles propagating intermediate build results between nodes? or is it that a pipeline runs on a single worker node and data isn't move around?

Do we know roughly how we'll structure the pipelines in terms of which types of tests would be run, and how? eg. unit, pact, e2e? It's totally fine if we don't know the answers to these questions now but it would be helpful to note what we do know and what we don't know yet.

rjbaker: I don't know exactly how we'll structure the pipelines at this point. But E2E tests should be able to fit in to the pipelines we have already built without too much issue.

rjbaker: Concourse is able to produce artefacts after a build stage. However the instance on which a build runs is ephemeral. For this reason Tech Ops have provided a "local" ECS repo to which candidate Docker images can be pushed for reuse elsewhere in a CI pipeline without needing to use Docker hub. Authentication is handled automatically. This provides similar functionality to locally tagging builds on a single Jenkins box, which wouldn't scale anyway if we used Jenkins worker nodes.

PR #130, opened 2020-03-13
PP-6254 Add deploy carbon-relay job

danworth:
We might want to move carbon-relay to its own repo to avoid it being
redeployed each time pay-omnibus is updated which may well not be
changing carbon-relay set-up.

### DIFF
--- a/ci/pipelines/deploy.yml
+++ b/ci/pipelines/deploy.yml
@@ -97,6 +97,10 @@ groups:
       - direct-debit-smoke-tests-staging
       - products-smoke-test-staging
 
+  - name: Carbon Relay
+    jobs:
+      - deploy-carbon-relay-staging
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy.yml
```diff
@@ -913,6 +917,17 @@ jobs:
           <<: *push-app-config
           path: artefact
 
+  - name: deploy-carbon-relay-staging
+    serial_groups: [carbon-relay-stg]
+    plan:
+      - get: omnibus
```
richardTowers: I think we could make this deploy a bit less often by putting a path filter on `omnibus` - at the moment we're going to deploy a new Carbon Relay for every change in the infra, which is a bit much.

Not a problem for now though.

PR #132, opened 2020-03-18
PP-6233: PaaS metric exporter infra

rjbaker:
Adds relevant CloudFoundry infra (via Terraform) and CI Concourse deployment tasks to deploy the PaaS metric-exporter application.

### DIFF
--- a/paas/metric-exporter/manifest.yml
+++ b/paas/metric-exporter/manifest.yml
@@ -1,0 +1,22 @@
+---
+applications:
+- name: metric-exporter
+  memory: 100M
+  instances: 1
+  stack: cflinuxfs3
+  buildpacks:
+    - go_buildpack
+    - https://github.com/alphagov/env-map-buildpack.git#v2
+  env:
+    GOPACKAGENAME: github.com/alphagov/paas-metric-exporter
+    # Provided via the metric-exporter-secret-service
+    API_ENDPOINT:
+    PASSWORD:
+    STATSD_ENDPOINT:
+    STATSD_PREFIX:
+    USERNAME:
+
+  health-check-type: process
+  no-route: true
+  services:
+    - metric-exporter-secret-service
--- a/paas/metric-exporter/manifest.yml
+++ b/paas/metric-exporter/manifest.yml
@@ -20,3 +20,4 @@ applications:
   no-route: true
   services:
     - metric-exporter-secret-service
+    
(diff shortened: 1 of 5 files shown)


paas/metric-exporter/manifest.yml
```diff
@@ -0,0 +1,23 @@
+---
+applications:
+- name: metric-exporter
+  memory: 100M
+  instances: 1
```
danworth: I think we're running 2 instances of carbon-relay. We haven't done any performance testing and we're leaving the fine tuning for later but the idea was to have multiple just to see how it behaves. Do you think we'll end up with more than on metric-exporter? If so, maybe add another instance?

terraform/staging-paas/terraform.tfvars
```diff
@@ -191,5 +191,18 @@ credentials = {
...
+  metric_exporter = {
+    pay_low_pass_secrets = {
+    }
+
+    static_values = {
+      cf_api_endpoint  = "https://api.cloud.service.gov.uk"
+      cf_username      = "placeholder"
+      cf_password      = "placeholder"
+      statsd_endpoint  = "statsd.hostedgraphite.com:8125"
+      statsd_prefix    = "placeholderapikey.paas.system"
```
danworth: I added the hosted graphite api key to pay-low-pass, see carbon-relay for an example:
https://github.com/alphagov/pay-omnibus/blob/d4471c0a1fc3bf4300f00daa7ca60127ede51b18/terraform/staging-paas/terraform.tfvars#L181

`hosted_graphite_api_key = "hosted_graphite/prod/api_key"`

paas/metric-exporter/manifest.yml
rjbaker: I'm almost certain it will double report if we run more than one instance. It has no state or shared buffer so both instances will be working independently and pulling the same metrics for a given time frame. Carbon relay is fine as it's essentially a forwarding proxy balanced via DNS and we're running multiple instances for HA and so that it can keep up with the metric volume from multiple app instances.

terraform/staging-paas/terraform.tfvars
rjbaker: Thank you 👍

paas/metric-exporter/manifest.yml
danworth: yeah that makes sense, good point 👍

PR #133, opened 2020-03-20
BAU: Trigger carbon-relay deploy only when app config changes

rjbaker:
Defines a separate resource for the carbon-relay app source by
using the "path" attribute of the git resource. This ensures
the deploy job is triggered only when a change is made to files
within the specified directory, rather than every commit to the
repo.

### DIFF
--- a/ci/pipelines/deploy.yml
+++ b/ci/pipelines/deploy.yml
@@ -153,6 +153,15 @@ resources:
           username: pay-deploy
           password: ((readonly_local_user_password))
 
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy.yml
```diff
@@ -153,6 +153,15 @@ resources:
           username: pay-deploy
           password: ((readonly_local_user_password))
 
+  - name: carbon-relay-source
```
danworth: Brilliant 👍

PR #143, opened 2020-04-16
PP-6320: Terraform support for CloudFoundry V3 API

rjbaker:
The v3 attribute has been added to a fork of the CloudFoundry Terraform provider. This ensures the V3 API is used when creating apps allowing use of "v3" commands, such as zdt app deployments and application sidecar definitions.

See: alphagov/terraform-provider-cf#1

To run the Terraform, the binary of the forked plugin must be built and installed into your local ~/.terraform.d/plugins directory.

https://www.terraform.io/docs/extend/how-terraform-works.html#plugin-locations


rorymalcolm:
notifications changes are good 👍

PR #154, opened 2020-04-24
PP-6371: Create PR pipeline

rjbaker:
**What is it?**
Adds a new pipeline for testing pull requests on all application repositories.

Uses the [pull request Concourse resource](https://github.com/telia-oss/github-pr-resource) to run unit and integration tests on Github PRs.

Java builds use a docker-in-docker container environment to allow Java Test Containers to pull and start Postgres and other Docker test dependencies within the test environment.

Within the pr.yml, the first PR build job definition specifies global `max_in_flight` (3 per repo) and `build_log_retention` (500 builds) parameters. These are inherited by all jobs but can be changed or overridden by subsequent job definitions as required.

### DIFF
--- a/ci/tasks/java-tests.yml
+++ b/ci/tasks/java-tests.yml
@@ -1,0 +1,40 @@
+container_limits: {}
+platform: linux
+
+image_resource:
+  type: registry-image
+  source:
+    repository: amidos/dcind
+
+caches:
+  - path: src/.m2
+
+inputs:
+  - name: src
+
+run:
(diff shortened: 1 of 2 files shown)


ci/tasks/java-tests.yml
```diff
@@ -0,0 +1,40 @@
+container_limits: {}
```
danworth, line 1: I think I committed this for future use, can swap in for the in-lined yaml task definition once we're happy with it.

PR #161, opened 2020-04-29
PP-6109: Add Publicapi WAF ACL config

rjbaker:
### What?
Adds a ruby script to create a basic WAF ACL using the WAFv2 API.
The ACL uses a JSON template to configure AWS managed rulesets (a new feature in WAFv2).
Composed into a Terrafrom module so it can be applied to other CloudFront distributions.

### How to test?
 - Run `terraform apply` in the aws staging environment
 - Check ACL and resource association in AWS console. Be sure to select *global* region to see ACLs applied to CloudFront distributions.
 - Matching rules are currently set to *count* not *block*, so we can measure stats.
 - The resource does not currently support updating WAF rule template in place.

The Terraform resource to support WAFv2 is due to be merged into the AWS provider.

See: https://github.com/terraform-providers/terraform-provider-aws/pull/12688

Co-authored-by: @rorymalcolm

### DIFF
--- a/terraform/modules/aws/waf_v2_acl/configure_aws_wafv2.rb
+++ b/terraform/modules/aws/waf_v2_acl/configure_aws_wafv2.rb
@@ -1,0 +1,41 @@
+#!/usr/bin/env ruby
+
+require 'json'
+require 'logger'
+
+def create_acl!(name, description, rules)
+  out = `aws wafv2 create-web-acl \
+        --scope CLOUDFRONT \
+        --region us-east-1 \
+        --name '#{name}' \
+        --description '#{description}' \
+        --default-action Allow={} \
+        --visibility-config SampledRequestsEnabled=true,CloudWatchMetricsEnabled=false,MetricName=#{name}AclMetrics \
+        --rules '#{rules}'`
+  JSON.load(out).fetch('Summary').fetch('Id')
+end
+
+def wafv2_acl_id(name)
+  JSON.load(`aws wafv2 list-web-acls --scope CLOUDFRONT --region us-east-1`)
+    .dig('WebACLs')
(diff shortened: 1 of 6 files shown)


terraform/modules/aws/waf_v2_acl/configure_aws_wafv2.rb
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env ruby
+
+require 'json'
+require 'logger'
+
+def create_acl!(name, description, rules)
+  out = `aws wafv2 create-web-acl \
+        --scope CLOUDFRONT \
+        --region us-east-1 \
```
danworth, line 9: Is there any benefit to making `region` a variable? I'm thinking in a DR scenario and we want to run this tf to stand up our environment elsewhere?

terraform/modules/aws/waf_v2_acl/rules.tpl
```diff
@@ -0,0 +1,74 @@
+${jsonencode([
```
danworth, line 1: Not sure if refining the rules is part of this but looks sensible, any idea how these rules compare with our existing or is that a future piece of work?

terraform/modules/aws/waf_v2_acl/configure_aws_wafv2.rb
rjbaker: It has to be us-east-1 since that's the "global" region for Cloudfront. Confusing, I know.

terraform/modules/aws/waf_v2_acl/rules.tpl
rjbaker: I've just chucked some basic rules in for now. They will need to be refined.

PR #178, opened 2020-05-13
PP-6448: WAF Log Kinesis Firehose Delivery Stream

rjbaker:
Deploys a Kinesis Firehose Delivery Stream used to process and deliver logs from WAF ACLs to Splunk.

- Uses a lambda function to process logs to required Splunk HEC format
- Drops "Allow" logs, so only Block and Count actions are delivered.
- Failed logs are delivered to an S3 bucket
- Adds log groups and streams for debugging Lambda function and Firehose Delivery
- Requires Splunk HEC Endpoint and Token as input variables

### DIFF
--- a/lambda/waf-to-splunk-kinesis-data-transformation/index.js
+++ b/lambda/waf-to-splunk-kinesis-data-transformation/index.js
@@ -1,10 +1,13 @@
 exports.handler = async (event, context) => {
     /* Process the list of records and transform them */
+    const okCount = 0
+    const droppedCount = 0
     const output = event.records.map((record) => {
         const eventJson = Buffer.from(record.data, 'base64').toString('utf8')
         const event = JSON.parse(eventJson)
 
         if (event.action === 'ALLOW') {
+            droppedCount++
             return {
                 recordId: record.recordId,
                 result: 'Dropped',
@@ -12,6 +15,7 @@ exports.handler = async (event, context) => {
             };
         }
 
+        okCount++
         const data = {
             time: (event.timestamp / 1000),
             host: "lambda",
@@ -27,6 +31,6 @@ exports.handler = async (event, context) => {
             data: Buffer.from(JSON.stringify(data), 'utf8').toString('base64')
         }
     });
-    console.log(`Processing completed. Successful records ${output.length}.`)
+    console.log(`Processing completed. Successful records ${output.length}. Dropped ${droppedCount}. Ok ${okCount}.`)
     return { records: output };
 };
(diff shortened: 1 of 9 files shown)


lambda/waf-to-splunk-kinesis-data-transformation/index.js
```diff
@@ -1,17 +1,21 @@
 exports.handler = async (event, context) => {
     /* Process the list of records and transform them */
+    const okCount = 0
+    const droppedCount = 0
```
mrlumbu, line 4: these aren't constants, should be `let`

terraform/modules/aws/waf_logging/kinesis-firehose.tf
```diff
@@ -1,5 +1,82 @@
...
+      processors {
+        type = "Lambda"
+
+        parameters {     
+            parameter_name  = "LambdaArn"
+            parameter_value = "${aws_lambda_function.kinesis_data_transformation_lambda.arn}:$LATEST"
+        }
+        parameters { 
+            parameter_name  = "BufferSizeInMBs"
+            parameter_value = 3
```
mrlumbu, line 31: any particular reason for this value (and one below)?

lambda/waf-to-splunk-kinesis-data-transformation/index.js
```diff
@@ -27,6 +31,6 @@ exports.handler = async (event, context) => {
             data: Buffer.from(JSON.stringify(data), 'utf8').toString('base64')
         }
     });
-    console.log(`Processing completed. Successful records ${output.length}.`)
+    console.log(`Processing completed. Successful records ${output.length}. Dropped ${droppedCount}. Ok ${okCount}.`)
```
danworth, line 34: I like you've added the `dropped` and `ok` counts 👍 Maybe change `Successful` to `Total` since it's a bit confused with the new `Ok` total?

rjbaker: All records are returned whether they're dropped or not. I'll update it be clearer.

lambda/waf-to-splunk-kinesis-data-transformation/index.js
rjbaker: Sorry, I originally had the counts in an object and updated the props.

terraform/modules/aws/waf_logging/kinesis-firehose.tf
rjbaker: Yes, the buffer size has to be below the maximum that a lambda can process. If the lambda is delivered a huge buffer it will fail. 60 seconds is the minimum, but is dynamically raised if delivery falls behind. It's just a mandatory property for some reason.

rjbaker: I'll document this 😄

PR #196, opened 2020-06-02
PP-6234 Create Users For Apps With SQS Access

rorymalcolm:
- Creates new users for the applications which will have access to SQS
and KMS for encypting queue entries


rorymalcolm:
Applied

rorymalcolm:
currently applied

PR #197, opened 2020-06-02
PP-6504 Consumer PR pact tests

mrlumbu:
Adds consumer PR pact tests for all consumers.

The general structure is:
- publish pacts in a separate task after the consumer pact tests are run,
  and tag pact with PR number
- in a separate job, first check whether any provider tests need to be
  run by calling pact-cli `can-i-deploy` - this avoids unnecessarily
  running provider verifications
- pass list of providers needing verification to provider verification tasks
  which runs all required provider verifications in parallel
- report back to PR on status of provider verification

This structure is more efficient and parallelised than current implementation,
and is fairly self explanatory when viewed in concourse.

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -172,6 +172,12 @@ definitions:
     task: test
     privileged: true
     file: omnibus/ci/tasks/node-build-pr.yml
+  - &publish-pacts
+    task: publish pacts
+    privileged: true
+    params:
+      consumer_name: updateThisValue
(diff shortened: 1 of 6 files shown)


ci/docker/concourse-runner/Dockerfile
```diff
@@ -13,6 +13,8 @@ RUN apk add --no-cache \
   maven \
   tini \
   docker-compose>1.25.4-r1 \
+  git \
+  curl \
```
danworth, line 17: we've got curl twice :)

PR #201, opened 2020-06-04
PP-6504 provider pact tests

mrlumbu:
Adds provider PR pact tests for all providers.

Modifies the `pact-check` task to be able to accommodate pact provider tests
being run in context of a consumer PR build (where only the tests between
that particular consumer and the provider are needed) and tests run in
context of provider build (where all provider tests need to be run against
the master tagged consumer pacts). The `pact-check` task is renamed to
pact-provider-test-preflight check to try to communicate its purpose.

Also modifies the `pact-provider-verification` task to make the interface tighter - it now only expects a folder with a bunch of files defining test config, and
a test target where it should run the provider tests.


ci/docker/concourse-runner/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN apk add --no-cache \
   maven \
   tini \
   docker-compose>1.25.4-r1 \
+  git \
```
rjbaker, line 16: It doesn't matter too much, but git is available from the alpine main repo as well as edge. This will install a bleeding edge version.

PR #203, opened 2020-06-09
PP-6234 Add Capture Queue Read/Write

rorymalcolm:
- Connector should be able to read and write from the capture queue.


terraform/modules/aws/card-connector.tf
```diff
@@ -33,6 +33,7 @@ resource "aws_iam_policy" "card_connector_queue_policy" {
           "Resource": [
             "${aws_sqs_queue.payout_reconcile.arn}",
             "${aws_sqs_queue.payment_event.arn}"
+            "${aws_sqs_queue.capture.arn}"
```
rjbaker, line 36: Needs a preceeding comma

PR #208, opened 2020-06-19
Pp 6676 trigger deploy from master merge

danworth:
Just trialling triggering deploy from master merge, adminusers first. Am I missing anything? Seems straight forward. Thought I'd seek some input before rolling it out for all apps. I've set the deploy pipeline from this change so you can see it working for Adminusers.

### DIFF
--- a/ci/pipelines/deploy.yml
+++ b/ci/pipelines/deploy.yml
@@ -162,89 +162,6 @@ resources:
       paths:
         - paas/carbon-relay/**
 
-  # Travis Builds
-  - &travis-build
-    name: card-frontend-ci-build
-    type: travis
-    icon: sync
-    source: &travis-build-source
-      repository: alphagov/pay-frontend
-      travis-token: ((travis-api-token))
-      branch: master
-      pro: true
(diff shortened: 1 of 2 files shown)


ci/tasks/checkout-commit.yml
```diff
@@ -8,6 +8,7 @@ image_resource:
 
 inputs:
   - name: build-info
+    optional: true
```
danworth, line 11: only whilst other apps are triggered by Travis and using build-info

ci/pipelines/deploy.yml
```diff
@@ -651,21 +645,23 @@ jobs:
...
       - get: omnibus
-      - get: build-info
-        resource: adminusers-ci-build
-        trigger: true
       - get: src
         resource: adminusers-source
-      - task: checkout-commit
+        trigger: true
+      - task: calculate-release-number
         file: omnibus/ci/tasks/checkout-commit.yml
```
danworth, line 652: will change take file name when all apps updated.

PR #212, opened 2020-06-21
Remove special DB migration command for adminusers

rhowe:
With a bit of luck, this command is going away - its job being done by the
standard 'db migrate' command. See https://github.com/alphagov/pay-adminusers/pull/780

It's only needed when starting from a blank database, so removing it shouldn't
affect any pre-existing environments.


katstevens:
Closing as we've removed this file 🗑️

PR #215, opened 2020-06-22
PP-6615 Send HG Metrics Regarding Smoke Tests

rorymalcolm:
- Send two new metrics to hosted graphite regarding smoke tests run
statuses

### DIFF
--- a/ci/tasks/send-hg-metric.yml
+++ b/ci/tasks/send-hg-metric.yml
@@ -1,0 +1,26 @@
+---
+platform: linux
+
+image_resource:
+  type: registry-image
+  source:
+    repository: alpine
+    tag: 3.12.0
+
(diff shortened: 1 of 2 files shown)


ci/tasks/send-hg-metric.yml
```diff
@@ -0,0 +1,26 @@
...
+  METRIC_NAME: metricname
+  METRIC_VALUE: metricvalue
+
+run:
+  path: bash
+  dir: src
+  args:
+    - -eu
+    - |
+        apk add netcat-openbsd
```
danworth, line 24: Does concourse re-use this container or is it installing netcat-openbsd every single time we send a metric? If so we might want to create an image with it already on and use that?

PR #216, opened 2020-06-22
PP-6615 Add Netcat

rorymalcolm:
- Add netcat-openbsd to the omnibus base image


rjbaker:
> Any idea why we run `RUN apk add --no-cache` twice? Shall we just combine them?

Because https://github.com/alphagov/pay-omnibus/pull/216/files#diff-2836f836bdee2c2a6c651424412b7d39L17

PR #224, opened 2020-06-24
BAU: Prevent builds triggered by external forks

rjbaker:
Prevents Concourse running PR builds on external, untrusted forks.

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -566,6 +566,7 @@ resources:
     icon: github-circle
     check_every: 1h
     source: &pull-request-source
+      disable_forks: true
       repository: alphagov/pay-connector
       access_token: ((github-access-token))
 


ci/pipelines/pr.yml
```diff
@@ -566,6 +566,7 @@ resources:
     icon: github-circle
     check_every: 1h
     source: &pull-request-source
+      disable_forks: true
```
rhowe, line 569: welp

rorymalcolm: 👋 byeeeee

rjbaker: See #225 Tests run in a container env with elevated permissions for 🐳 in 🐳  reasons.

rhowe: I might be able to make that mostly go away... no promises though

PR #236, opened 2020-06-26
PP-6704 Run provider pact validation on master merge

danworth:
When a provider is merged into master run pact validation using the
commit one before the merge commit which will be the commit from the
head of the PR. The pact broker will know about this commit since it
will have been tested during the PR build leading up to this merge. If the
pacts with its consumers are still valid then exit. If not then rerun
the validation for the consumer. If the rerun fails then manual
intervention will be necessary as it would be today with Jenkins CI.

**HOW**
This has been set to our deploy pipeline and it looks good so far, for example 
https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy/jobs/build-card-connector/builds/25

Also we ran through the plan with Oz and he was onboard. I think we can bring together some of the pact verification scripts but at present there are subtle differences that mean they can't be reused as is.

### DIFF
--- a/ci/pipelines/deploy.yml
+++ b/ci/pipelines/deploy.yml
@@ -1,5 +1,5 @@
 ---
-definitions: 
+definitions:
   - &notify-hg
     task: change-this-value
     file: omnibus/ci/tasks/send-hg-metric.yml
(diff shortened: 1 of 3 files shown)


ci/pipelines/deploy.yml
```diff
@@ -524,6 +530,27 @@ jobs:
       - get: src
         resource: card-connector-source
         trigger: true
+      - in_parallel:
```
rjbaker, line 533: Nice 👍

ci/tasks/pact-provider-master-verification.yml
```diff
@@ -0,0 +1,48 @@
...
+        echo "CLEANUP TRIGGERED"
+        clean_docker
+        stop_docker
+        echo "CLEANUP COMPLETE"
+      }
+
+      trap cleanup EXIT
+      source /docker-helpers.sh
+
+      pr_sha="$(git rev-list --no-merges -n 1 HEAD)"
```
rjbaker, line 28: I didn't know `--no-merges` was a thing 👍

PR #259, opened 2020-07-22
BAU Fix pact broker host var when running mvn test

danworth:
The java pact tests require a host rather than url for the pact broker
so remove the `https://` from the beginning.

## WHAT ##
Tested this on a hijacked container and it works.


danworth:
Regarding the failed pipe clean check, this is unrelated to the file I've changed which is clean:
`FILE ci/tasks/pact-provider-master-verification.yml (task)`

PR #267, opened 2020-07-27
PP-6845 Add Chromium

rorymalcolm:
- Karma tests need chromium in the container


ci/tasks/node-build-pr.yml
```diff
@@ -31,7 +31,9 @@ run:
       apt-get update
       apt-get -y install ca-certificates wget python libgtk2.0-0 libgtk-3-0 libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb
 
-      apt-get -y -t stretch-backports install git 
+      apt-get -y -t stretch-backports install git chromium
+
+      export CHROME_BIN='/usr/bin/chromium'
```
rjbaker, line 36: This container is used for all node/js build tasks. This will cause chromium to be downloaded and installed every time one a node job is run. js-commons could probably do with its own task.

PR #291, opened 2020-08-20
PP-6944 Add pr build time job to each app

danworth:
Record the time a pr takes to successfully build. Since e2e is optional
this is not a pre-requisite for measuring the time and is therefore not
included in the duration. We may wish to consider this later on.


rorymalcolm (approved):
LGTM!

PR #296, opened 2020-08-21
PP-6914: Pin to TF 0.12 compatible cyber sec module

rjbaker:
Update cybersec audit module to a TF 0.12 compatible version


rorymalcolm:
This has been done in a future PR and can be closed

PR #298, opened 2020-08-25
PP-6944 Store Dashboard JSON

rorymalcolm:
- When terraform is run on grafana it will destroy the state of our
dashboards, we should keep a record of their state in JSON so we can restore
from the existing setup


danworth (approved):
Nice 👍

PR #300, opened 2020-08-26
PP-6914: Add VPC to AWS environments

rjbaker:
### What is it?

 - Adds a default VPC with relevant configuration.
 - Allows VPC CIDR range to configured per-environment to support VPC peering without overlapping ranges.
 - Configures the default security group with null settings
 - Adds a default route table

 - Various formatting fixes 🧹


danworth (approved):
Looks good 👍

PR #305, opened 2020-08-28
PP-6914: Add RDS subnets and PaaS VPC ingress

rjbaker:
- Creates a subnet in each AZ for placement of RDS instances
 - CIDR for staging VPC is `172.20.0.0/16`. 
 - RDS subnet range starts at 0, (defined in `subnet_ranges`). This produces the following subnets: `172.20.0.0/24`, `172.20.1.0/24`, `172.20.2.0/24`.
 - Creates a security group with a rule allowing ingress from PaaS Ireland VPC internal (peered) CIDR range

### DIFF
--- a/terraform/modules/aws/rds.tf
+++ b/terraform/modules/aws/rds.tf
@@ -1,0 +1,11 @@
+data "aws_availability_zones" "available" {}
+
+module "rds" {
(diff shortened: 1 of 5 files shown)


danworth (approved):
nice 👍

PR #307, opened 2020-08-28
PP-6913: Provision RDS instances

rjbaker:
Provisions all RDS instances within an RDS subnet group in a given environment.

- Accepts a map of database instances with optional configuration
- Default configuration is applied where values are omitted
- Configures a VPC peering connection to the RDS subnet group from GOV.UK PaaS
- Adds peering connection to VPC routing table


rjbaker (commented):
Minor comment about the vpc peering id, but this can be refactored in a separate PR.

danworth (approved):
Thanks for the comment Richard, will add it into the next pr which will soon follow. Approving this since we've all worked on it at some point and its been applied and works.

PR #308, opened 2020-09-01
PP-6914: Update VPC cidr & RDS base subnet

rjbaker:
This change updates the Staging AWS account VPC CIDR to 172.16.0.0/16 and updates the base RDS subnet group range to a CIDR within a range compatible with PaaS for VPC peering.


rorymalcolm (approved):
LGTM, updated subnet and added base range are good 👍

PR #310, opened 2020-09-02
PP-7062 Add serial groups to pr pipeline

danworth:
We still exhaust worker memory when multiple prs are raised across
applications despite turning off e2e tests. We want to get the pipeline
running reliably and improve performance from there if necessary. This
introduces `serial_groups` which limits all jobs tagged with the same
`serial_groups` to run one at a time. To start with each type of test
e.g. unit, integration, pact has its own serial group tag.

We should monitor the effect this has upon our PR pipeline SLOs, both build
time and error rate.

## WHAT ##
After another chat with Chris Farmiloe we've decided to implement `serial_groups`.

Mostly as described. I opted for `node_test` as a separate tag since the way we test node apps includes both unit, integration and Cypress within the same job so they don't really fall into a more granular category like with java apps. Alternatively we could use `serial_groups: [unit_test, integration_test, cypress_test]` for the node apps but that seems like it might disproportionately impact node applications - still we can adopt that if necessary.

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -727,6 +727,7 @@ resources:
 jobs:
   - <<: *job-definition
     name: card-connector-unit-test
+    serial_groups: [unit-test]
     plan:
     - <<: *get-pull-request
       resource: card-connector-pull-request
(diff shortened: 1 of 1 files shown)


rorymalcolm (approved):
LGTM 👍

PR #312, opened 2020-09-04
Pp 6913 add remaining dbs

danworth:
- Add remaining dbs in similar fashion to `card-connector`.
- Fix up deprecation warnings.
- Ignore changes to snapshot identifiers and remove them now that we've created the databases since we don't want to reset the dbs to these snapshots each time we TF the environment.
- Pass the paas vpc peering connection name in from `staging-aws/site.tf` since it'll be different for each environment.

### DIFF
--- a/terraform/modules/aws/rds/main.tf
+++ b/terraform/modules/aws/rds/main.tf
@@ -18,6 +18,7 @@ resource "aws_db_instance" "postgres" {
 
   allocated_storage   = try(each.value.allocated_storage, var.default_rds_params.allocated_storage)
   storage_type        = try(each.value.storage_type, var.default_rds_params.storage_type)
+  storage_encrypted   = true
   engine              = try(each.value.engine, var.default_rds_params.engine)
   engine_version      = try(each.value.engine_version, var.default_rds_params.engine_version)
   instance_class      = try(each.value.instance_class, var.default_rds_params.instance_class)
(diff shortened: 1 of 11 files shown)


rorymalcolm:
this is current applied btw

terraform/modules/aws/rds/main.tf
```diff
@@ -18,6 +18,7 @@ resource "aws_db_instance" "postgres" {
 
   allocated_storage   = try(each.value.allocated_storage, var.default_rds_params.allocated_storage)
   storage_type        = try(each.value.storage_type, var.default_rds_params.storage_type)
+  storage_encrypted   = true
```
rjbaker, line 21: Good catch 👍

terraform/staging-aws/site.tf
```diff
@@ -49,11 +49,12 @@ module "staging" {
...
-  environment   = "staging"
-  domain_name   = "gdspay.uk"
-  paas_domain   = "cloudapps.digital"
-  vpc_cidr      = "172.16.0.0/16"
-  rds_instances = var.rds_instances
+  environment           = "staging"
+  domain_name           = "gdspay.uk"
+  paas_domain           = "cloudapps.digital"
+  vpc_cidr              = "172.16.0.0/16"
+  paas_vpc_peering_name = "pcx-0d726a5057505ab0b"
```
rjbaker, line 56: 👍

rjbaker (approved):
Looks great 👍

PR #313, opened 2020-09-04
BAU Recursive Terraform Format

rorymalcolm:
- Run terraform fmt on everything we've got to remove any future
git hook blocks


danworth (approved):
Nice clean up auto fmt 👍

PR #314, opened 2020-09-04
BAU: Enable RDS performance insights with default settings for all DBs

rjbaker:
Enables performance insights to aid debugging of DB performance issues.

- Provides 7-day retention of performance data.
- Performance data is encrypted with default RDS KMS key

More info: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_PerfInsights.Overview.html


danworth (approved):
ah nice 👍

PR #315, opened 2020-09-04
PP-7020 Add db credentials for card-connector

danworth:
Add the peered in rds credentials for connector db to the
card-connector-secret-service. The `db_host` comes from a data
`aws_db_instance`. This can be extended using a `for_each` to get the
db_instances for the other apps.


rjbaker (approved):
Looks good. Appears to apply credentials correctly. Requires env map config to be updated for apps with databases.

PR #316, opened 2020-09-04
PP-7020: Add db SSL connection option to static app configuration

rjbaker:
This option is used by apps to ensure DB connections are made over TLS.

A corresponding change should be made to the env-map.yml file in each app's repository to ensure this option is mapped to the `DB_SSL_OPTION` environment variable from the secret service at run time.


danworth (approved):
👍

PR #318, opened 2020-09-08
BAU: Remove DD frontend & DD connector infrastructure

rjbaker:
Direct Debit is being mothballed for a while. Remove PaaS & related AWS infrastructure related to both DD frontend and DD connector applications.


terraform/staging-account/tfstate.json
```diff
@@ -0,0 +1,485 @@
+{
```
danworth, line 1: Is the state file supposed to be included?

danworth (approved):
Looks good 🗑️  ✔️

terraform/staging-account/tfstate.json
rjbaker: Removed 👍

PR #320, opened 2020-09-11
PP-7025 Use separate hosted zones for sub domain delegation

danworth:
Currently the staging account has the domain registration for
`gdspay.uk`. To make the terraform consistent for any environment
(test, staging etc) create a hosted zone in which the A records for that
environments' apps are created in. Delegate the subdomain to the new  hosted
zone from the hosted zone in staging which owns the `gdspay.uk`
registration; this is to be done manually for now by adding the
necessary name server record pointing to the name servers of the new
hosted zone.

Remove subdomain interpolation in favour of using `var.domain_name`
passed in from the environment module, e.g. `staging.gdspay.uk`..

Co-authored-by: dan worth <dan.worth@digital.cabinet-office.gov.uk>

## WHAT ##
We've applied this to staging and it all seems to work, smoke tests green and dns resolves as expected.

### DIFF
--- a/terraform/modules/aws/card-frontend.tf
+++ b/terraform/modules/aws/card-frontend.tf
@@ -1,6 +1,6 @@
 resource "aws_route53_record" "card_frontend" {
-  zone_id = data.aws_route53_zone.root.zone_id
-  name    = "card-frontend.${local.subdomain}"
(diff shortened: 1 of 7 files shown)


rorymalcolm (approved):
LGTM!

PR #321, opened 2020-09-11
PP-7025 adding test environment

rorymalcolm:
- Add test environment configuration

- Fix VPC nullification

- Use `null` instead of `""`

Co-authored-by: dan worth <dan.worth@digital.cabinet-office.gov.uk>


danworth (approved):
We should refactor out the pass store in `secrets` directory in favour of moving these to `pay-low-pass`.

PR #322, opened 2020-09-11
PP-7025 Add rds instances for test environment

danworth:
Add configuration for rds postgres instances in the test environment.
The sizes accommodate the snapshots taken from test-12 with a little
extra head room.

Co-authored-by: rory malcolm
<rory.malcolm@digital.cabinet-office.gov.uk>

## WHAT ##
This has been applied first using snapshots taken from test-12 and the the `snapshot_identifier` removed from the config to prevent it rolling back each the terraform is run. Similar approach as per staging.


rorymalcolm (approved):
LGTM!

PR #323, opened 2020-09-13
PP-7150: Migrate AWS WAF definitions to use new WAFv2 Terraform resource

rjbaker:
Removes `null_resource` and Ruby definition script for AWS WAF v2 in favour of native Terraform support now the `aws_wafv2_web_acl` resource is available.

WAF ACL definitions are deliberately declarative rather than modular as these will likely require refinement resulting in different rules for each public service. This would require unnecessarily complex ACL config and logging rules to be passed to such a module.

As written, each application's ACL may be updated independently without causing side effects for other apps or accidental misconfiguration.


danworth (approved):
🥳  this is great and on a Sunday evening too 😄 . I like that it'll be much easier to find this terraform for each app since its part of the app's main terraform 👍 . Am I right in thinking if there was any environment specific ACL config, like permitted IPs for Notifications then it could be passed as a var into the aws module (and if no var then a default applied in its place)?

rjbaker:
> 🥳 this is great and on a Sunday evening too 😄 . I like that it'll be much easier to find this terraform for each app since its part of the app's main terraform 👍 . Am I right in thinking if there was any environment specific ACL config, like permitted IPs for Notifications then it could be passed as a var into the aws module (and if no var then a default applied in its place)?

Yes spot on ☝️  I was thinking about this earlier. We could probably do with a private repo in which to keep special TF modules containing the IPs and any other sensitive config etc. They can be pulled in and passed to the Notifications WAF definition.

rorymalcolm (approved):
LGTM

PR #324, opened 2020-09-14
PP-7154 Create Test Environment

rorymalcolm:
- Add PaaS test environment application skeleton with ancilliary
services and network policies.

### DIFF
--- a/terraform/test-paas/terraform.tfvars
+++ b/terraform/test-paas/terraform.tfvars
@@ -1,0 +1,241 @@
+// these are copied from staging-paas/terraform.tfvars and adapted for the
+// test environment, as the maturity of the secrets solution for PaaS develops
+// we will move to their secrets solution instead of this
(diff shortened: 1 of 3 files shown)


terraform/test-paas/terraform.tfvars
```diff
@@ -0,0 +1,238 @@
+credentials = {
```
danworth, line 1: Maybe worth noting that we decided to simply copy this `tfvars` file and adjust it for the test environment rather than refactoring and using some string interpolation for the environment because this is all going to get a major refactor when we get PaaS Secrets Solution, so it didn't seem worth spending time on any refactoring now.

rorymalcolm: Yeah agreed - will do this now

rjbaker (approved):
LGTM

PR #325, opened 2020-09-15
PP-7150 Create VPC Peering Connection Accepter

rorymalcolm:
- Create VPC peering connection accepter and prevent destruction of the
VPC peering connection


rjbaker (approved):
LGTM. Paired with @rorymalcolm to import the VPC peering connection into Terraform state.

PR #326, opened 2020-09-15
PP-7150 Remove Existing ACL Config

rorymalcolm:
- Remove the existing config for the ACL from the application terraform


rjbaker (approved):
LGTM. We will replace with TF WAF v2 resources in a subsequent PR.

PR #327, opened 2020-09-15
PP-7150: Web ACL Access Configuration

rjbaker:
`web_acl_id` property of an `aws_cloudfront_distribution` actually needs to be an ARN even though the docs state ID. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#web_acl_id

`aws_wafv2_web_acl_logging_configuration` need to use the `us` (us-east-1) provider because they are "global" resources.


rorymalcolm (approved):
LGTM (Lol)

PR #330, opened 2020-09-21
PP-7027 Deploy Test Smoke Test

rorymalcolm:
- Declare the smoke-tests-test application

### DIFF
--- a/paas/smoke-test-apps.yml
+++ b/paas/smoke-test-apps.yml
@@ -13,7 +13,7 @@ applications:
     - route: selenium-hub.apps.internal
 
   - &smoke-test-app
-    name: smoke-tests-staging
+    name: smoke-tests-((environment))
     docker:
       image: govukpay/endtoend:latest-master
     health-check-type: process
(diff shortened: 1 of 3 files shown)


paas/smoke-test-apps.yml
```diff
@@ -34,5 +34,8 @@ applications:
       NOTIFICATIONS_URL: ((notifications_url))
       DIRECT_DEBIT_GOCARDLESS_WEBHOOK_SECRET: ((direct_debit_gocardless_webhook_secret))
 
+  - <<: *smoke-test-app
```
danworth, line 37: Not sure you can use the anchor as-is because the app urls differ across environment, e.g. `PUBLICAPI_URL` and `NOTIFICATIONS_URL`. Maybe you could do something like the following making sure to update the env files accordingly https://github.com/alphagov/pay-omnibus/tree/master/paas/env_variables
```
NOTIFICATIONS_URL: https://notifications.((environment))
```

danworth, line 37: Kind of think this could be one app definition with 
`name: smoke-tests-((environment))` couldn't it? We don't push the entire manifest of all apps, only ever one referenced by name? https://github.com/alphagov/pay-omnibus/blob/master/ci/pipelines/deploy.yml#L1034

danworth (approved):
Looks good 🚢

PR #332, opened 2020-09-21
PP-7024 VPC Peering Acceptance

rorymalcolm:
- Create vpc_peering_connection_accepter
- Accept connection
- Modify WAF creation rules

### DIFF
--- a/terraform/modules/aws/card-frontend.tf
+++ b/terraform/modules/aws/card-frontend.tf
@@ -113,6 +113,12 @@ resource "aws_wafv2_web_acl" "card_frontend" {
     allow {}
   }
 
+  lifecycle {
+    ignore_changes = [
+      rule,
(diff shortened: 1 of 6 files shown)


terraform/modules/aws/card-frontend.tf
```diff
@@ -113,6 +113,12 @@ resource "aws_wafv2_web_acl" "card_frontend" {
     allow {}
   }
 
+  lifecycle {
```
rjbaker, line 116: Are these meant to be in this PR?

rorymalcolm: Yeah, it reflects the current state - I was going to do them in a separate PR but I applied it all at once, I'm fine to remove it though

rjbaker (approved):

PR #334, opened 2020-09-22
PP-7026 Add space to cardid data deployment

danworth:
Possibly something changed since we last deployed it in April but need
to add the space and mapping from the `cardid-data` source to the
generic `artefact` directory which is used by the deploy script.

## WHAT ##
Sorts out deployment of cardid-data in test and staging. I've applied this and both successfully deploy, as do the corresponding cardid apps in both environments.


rorymalcolm (approved):
lgtm

PR #336, opened 2020-09-23
PP-7022 Fix can-i-deploy check

danworth:
Switch to `json` output which is easier to parse. The preivous logic was
not correctly identifying providers that required validation when they
had not yet been validated against the version of the contract in the
request. This uses a new container which adds `jq` to the
`pactfoundation/pact-cli` image we were already using.

## WHAT ##
Previous we were parsing the `can-i-deploy` table response with `awk`, we didn't know `json` format was available. This hopefully makes it clearer and correctly identifies cases when provider validation is required.

The bug in this script is mitigated because Jenkins has been running pact validation on main builds.

I have built and pushed the `govukpay/pact_with_jq` image to dockerhub: https://hub.docker.com/repository/docker/govukpay/pact_with_jq

Tested the `jq` with recent pr which highlighted this issue: https://github.com/alphagov/pay-publicapi/pull/1008
```
/pact # echo "$can_deploy" | jq '.matrix[] | select(.verificationResult.success != true) | .provider.name'
"ledger"
```

### DIFF
--- a/ci/docker/pact_with_jq/Dockerfile
+++ b/ci/docker/pact_with_jq/Dockerfile
@@ -1,0 +1,2 @@
+FROM pactfoundation/pact-cli
+RUN apk add jq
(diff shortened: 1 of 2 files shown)


rorymalcolm (approved):
bit easier to parse 🤪

PR #339, opened 2020-09-24
PP-7021 Create paas test releases

danworth:
After a pre-release with the tag format `paas_release-n` has been
deployed and the smoke tests have passed, create a pre-release with the tag
`paas_release_test-n` which can then be deployed into staging.

This is just for Adminusers and will apply the same to the other apps if
we agree on this approach.

## WHAT ##
As described above. I've tested this out and the resulting pre-release is here: 
https://github.com/alphagov/pay-adminusers/releases/tag/paas_release_test-62

### DIFF
--- a/ci/pipelines/test-deploy.yml
+++ b/ci/pipelines/test-deploy.yml
@@ -9,8 +9,8 @@ definitions:
       METRIC_NAME: change-this-value
       METRIC_VALUE: 0
 
-  - &create-release-metatdata
-    task: create-release-metatdata
+  - &create-release-metadata
+    task: create-release-metadata
     config:
       platform: linux
       image_resource:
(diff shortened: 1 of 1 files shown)


danworth:
To anyone reviewing this from beyond Hufflepuff, the `concourse-ci/test` always fail on the pipeline linter, we're yet to clean them up

ci/pipelines/test-deploy.yml
```diff
@@ -8,63 +8,108 @@ definitions:
       HOSTED_GRAPHITE_API_KEY: change-this-value
       METRIC_NAME: change-this-value
       METRIC_VALUE: 0
+
+  - &create-release-metatdata
```
alexbishop1, line 12: “metatdata”

alexbishop1 (changes requested):

alexbishop1 (approved):

ci/pipelines/test-deploy.yml
danworth: thanks I've fixed it

PR #340, opened 2020-09-24
PP-7128: Bootstrap performance testing AWS account

rjbaker:
Adds a Terraform module to manage state bucket and IAM roles allowing access to performance testing AWS account.


Krenair:
This is now applied, I'll send you the account ID when we've enabled root MFA.

katstevens:
We no longer need this AWS account (for performance testing the Pay PaaS environment). We don't think the team is able to access this account ourselves to see what's in the account. @rjbaker @Krenair would it be best to just close this PR and delete the account from the RE side?

mrlumbu:
Confirmed there's nothing in this account. We will need to close it when that is possible

PR #341, opened 2020-09-24
PP-7027 Remove smoke test user setup

danworth:
In the early days of the PaaS migration we used a bit of the `pay local`
cli to create a user to be used for the sandbox smoke tests. Now we are
using snapshots of RDS instances from our existing environments e.g.
`test-12` we want to use the existing smoke test services and accounts.

Therefore remove the concourse job that would create a new smoke test
user and add a simple script to help set the necessary values.

### DIFF
--- a/ci/pipelines/deploy.yml
+++ b/ci/pipelines/deploy.yml
@@ -111,7 +111,6 @@ groups:
       - deploy-selenium-hub
       - deploy-smoke-tests-staging
       - smoke-tests-staging-network-policies
-      - smoke-test-user-staging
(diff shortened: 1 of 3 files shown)


rjbaker (approved):
Looks good for more permanent environments. We can revisit the creation of dynamic smoke test users & creds for dev accounts later if required.

danworth:
Thanks Richard, yeah we'll need to consider temp environments and whether we'd have smoke tests running on them I guess; but definitely some convenient way to create a user to play around with 👍

PR #345, opened 2020-09-28
PP-7213 Run Publicapi's provider verification after its pushed pacts

danworth:
Publiapi is a java consumer. This change ensures that its provider
verification (where we run it's associated providers against it's latest
contracts) happens after it's pushed the latest contracts as part of the
`publicapi-integration-test` job.

with @SandorArpa

## WHAT ##
I don't think the other apps need to change. the node consumers push their pacts as part of their `test` job which is already a prerequisite for the pact verification. Ledger is more complicated being both consumer and provider and it's pipeline looks fine.

You can see the relationships here: https://pay-concourse-pact-broker.cloudapps.digital/

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -947,7 +947,7 @@ jobs:
     plan:
       - <<: *get-pull-request
         resource: publicapi-pull-request
-        passed: [publicapi-unit-test]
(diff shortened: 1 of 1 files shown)


SandorArpa (approved):
LOKTM
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #348, opened 2020-10-09
PP-7121 Tag connector pacts with main

danworth:
When we build connector after a merge to the main branch tag the pacts
for that version of connector with `main`.

This is new since Connector has just become a consumer.

## WHAT ##
Applied this to main pipeline and it works as expected, here are the connector contracts tagged with `main`
https://pay-concourse-pact-broker.cloudapps.digital/hal-browser/browser.html#/pacticipants/connector/versions/074423d2109c1575b918e122d5a4f1a9cfd3ddf6

### DIFF
--- a/ci/pipelines/main.yml
+++ b/ci/pipelines/main.yml
@@ -23,6 +23,16 @@ definitions:
         - build-output/*
       commitish: src/.git/ref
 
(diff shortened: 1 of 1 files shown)


stephencdaly (approved):
Looks good

PR #349, opened 2020-10-09
PP-7253 Validate pacts against consumers with main tag

stephencdaly:
We are now tagging pacts for commits when they have been merged to master with the "main" tag rather than "master". Update PRs to ensure we are validating providers against the correct consumer version.


danworth (approved):
Looks good 👍

PR #350, opened 2020-10-20
We were using the 'dev' tag for the pr resource

danworth:
Use the 'latest' tag image rather than 'dev'. I think we experimented
with the 'dev' tag because it had some functionality we were exploring
during the development. I don't think we're reliant upon it now, but the
'dev' tag is broken so we need to move from it.

## WHAT ##
Applied to concourse and this resolves the pr resource errors we suddenly started to see.


rorymalcolm (approved):
Looks likely not to start an incident to me

PR #352, opened 2020-10-28
PP-6761 Downgrade Concourse Github PR Resource

rorymalcolm:
- Downgrade the version of the Concourse Github PR Resource to see if it
is in fact new changes introduced to the resource check regarding PR
filtering that is causing the rate limiting error we are seeing.

### DIFF
--- a/ci/pipelines/main.yml
+++ b/ci/pipelines/main.yml
@@ -43,7 +43,7 @@ resource_types:
     type: registry-image
     source:
       repository: teliaoss/github-pr-resource
-      tag: dev
+      tag: v0.21.0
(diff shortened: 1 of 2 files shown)


ci/pipelines/main.yml
```diff
@@ -43,7 +43,7 @@ resource_types:
     type: registry-image
     source:
       repository: teliaoss/github-pr-resource
-      tag: dev
```
danworth, line 46: 🤮  lets do a check for any more of these uses of `dev` 😄

danworth (approved):

PR #353, opened 2020-11-25
PP-7336 Add new Connector env var for Stripe Ips

danworth:
Connector won't start without `STRIPE_ALLOWED_CIDRS` set so set it for
the docker-compose end-to-end setup.

## WHAT
Tested on hijacked container and this gets connector started and running.


nimalank7 (approved):
LGTM!

PR #355, opened 2021-01-20
PP-7652 Remove everything except pr pipeline

mrlumbu:
Removes all terraform, including some that corresponds to infrastructure
that hasn't been removed yet (mainly IAM roles for access to defunct accounts)
but I figure we can click ops them into oblivion when needed.

Removes the entire omnibus docs pipeline as that does not seem
to be a thing we use these days.

Removes entire paas dir as not needed.

I guess this purge might be too extreme, but as the old saying goes,
"better to delete too much than too little". Or "it's always in git"


katstevens:
I am extremely envious of the +/- ratio.

A couple of questions:
- Do we want to capture the deleted ADRs elsewhere (Google Drive, maybe)?
- What's happening with the `ci/tasks/endtoend/docker-config/ssl/make-selfsigned.sh` permissions change?
- Should we update the README to make clear which bits are left?

mrlumbu:
After discussing with @danworth I've reverted deleting a bunch of stuff out of CI and secrets. Dan is going to do some of that, and can finesse the rest but probably needs a bit more than I showed. Yes, moving adrs to google drive is a good idea - will do that. I've updated the README to be accurate. I plan do another pR to rename and be clearer on the new purpose of this repository.

mrlumbu:
All adrs are now here https://drive.google.com/drive/u/0/folders/1dWxfFMD5qyb_DPT1ufYqsFo9J4IrzZ95

katstevens (approved):
My favourite kind of PR 🔥

PR #357, opened 2021-01-21
PP-7652 Rename omnibus to ci

mrlumbu:
So that people know what this repo is for

### DIFF
--- a/ci/pipelines/docs.yml
+++ b/ci/pipelines/docs.yml
@@ -6,11 +6,11 @@ resource_types:
       repository: nulldriver/cf-cli-resource
 
 resources:
-  - name: omnibus-docs
+  - name: ci-docs
     type: git
     icon: github
     source:
-      uri: https://github.com/alphagov/pay-omnibus
+      uri: https://github.com/alphagov/pay-ci
       branch: master
       paths:
         - docs
@@ -29,7 +29,7 @@ jobs:
   - name: deploy
     serial: true
     plan:
-      - get: omnibus-docs
+      - get: ci-docs
         trigger: true
       - task: build
         config:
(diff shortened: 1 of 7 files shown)


ci/pipelines/docs.yml
```diff
@@ -39,13 +39,13 @@ jobs:
...
+            - name: ci-docs
           outputs:
-            - name: omnibus-docs/docs/build
+            - name: ci-docs/docs/build
           caches:
-            - path: omnibus-docs/docs/vendor
+            - path: ci-docs/docs/vendor
           run:
-            dir: omnibus-docs/docs
+            dir: ci-docs/docs
```
katstevens, line 48: Do we want to change folder names (etc) in the apps themselves?

mrlumbu: this references the `ci-docs` above which is an alias for this repo in the context of this pipeline. So think it needs to change no?

katstevens: OK - I don't know much about the tech docs app but I'm guessing if we merge this, then these changes will take effect on the next tech docs merge/deploy?

mrlumbu: yes this is a tricky change - we need to change the thing (the name of the repo) and references to the thing (contained in the configuration of the  pipelines in concourse). I have changed the the name of the repo. The changes in this PR change the references (including downstream aliases that don't really need to be changed but I have for consistency). The references will only be changed on concourse once `fly set-pipeline ...` is run (either by us manually, or by concourse which in a Inceptionesque twist can use this config to update itself). So in answer to your question, the tech-docs pipeline will be changed when we apply the changes in this branch. But tech docs repo itself should need no changes. Not sure if that's clear at all?

mrlumbu: I have applied the changes just to the concourse runner pipeline to check it works with the renaming - see here https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/concourse-runner/jobs/build-and-push/builds/13

mrlumbu: Screenshot 2021-01-21 at 16 41 58

katstevens: Excellent 👍

katstevens (approved):
Cool - have grepped the branch and no more `omnibus`es are lurking 🔥

PR #358, opened 2021-01-25
PP-7502 Delete docs pipeline config

danworth:
the `docs` pipeline was for our omnibus/paas documentation which is no
longer neeeded.


katstevens:
Where do we keep the pipeline config for deploying the Pay tech docs and the Pay team manual? I can't seem to find them...

danworth:
> Where do we keep the pipeline config for deploying the Pay tech docs and the Pay team manual? I assumed they were in this repo but I can't seem to find them...

The config for those is in the repo it is deploying. 
https://github.com/alphagov/pay-team-manual/blob/master/ci/pipeline/pay-team-manual.yml
https://github.com/alphagov/pay-tech-docs/blob/master/ci/pipeline.yml

For the ci pipelines which are common to various repos/apps then they're in here (arguably the two pipelines above are classed as ci related). I think we ought to formalise that decision in the team manual?

katstevens:
Gotcha 👍 I think we could put it on the READMEs for those repos? I'm happy to do that.

katstevens (approved):

PR #368, opened 2021-02-15
BAU: fix ci-pr-test failures

katstevens:
## What
The `pr-ci/ci-pr-test` job has been failing consistently for some weeks (see [an example failed build](https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/pr-ci/jobs/ci-pr-test/builds/47.1)).

This job has been using [pipecleaner](https://github.com/alphagov/paas-cf/tree/main/tools/pipecleaner#features) to run several checks:
- Concourse's own validation
- Shellcheck on any shell or bash scripts
- Resource check, to flag any unused resources in a job description
- (Pipecleaner can also run Rubocop, but we've disabled that option)

I don't think the repo-renaming itself caused the failures, however we can't check further back as the old `omnibus-pr-test` pipeline no longer exists on Big Concourse.

## How
I've gone by the shellcheck suggestions, and validated those using the [online shellcheck tool](https://www.shellcheck.net/), as my local version of `pipecleaner` wasn't working for some reason. 

I've used `bash` instead of `sh` to resolve most of the errors - we're using for some tasks already, so may as well be consistent.

I added two 'ignores' for shellcheck, in `pr.yml` and `tasks/publish-pacts.yml` - not because they are unfixable, but because they are just time-consuming enough to get bogged down in, and I suspect they're actually are fine in context. I am willing to be persuaded!

## Not included in this PR
I haven't fixed the deprecation warnings - there are a lot of them and this PR is already full of small fiddly changes.

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -1058,7 +1058,6 @@ jobs:
     plan:
     - <<: *get-pull-request
       resource: java-commons-pull-request
-    - <<: *get-ci
     - <<: *put-integration-test-pending-status
       put: java-commons-pull-request
     - task: test
@@ -2081,7 +2080,6 @@ jobs:
     plan:
     - <<: *get-pull-request
       resource: stubs-pull-request
-    - <<: *get-ci
     - <<: *put-test-pending-status
       put: stubs-pull-request
     - task: run tests
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -1386,11 +1386,12 @@ jobs:
         outputs:
           - name: pacts
         run:
-          path: sh
+          path: bash
           dir: src
           args:
             - -ec
             - |
+              # shellcheck disable=SC1091
               source /docker-helpers.sh
               start_docker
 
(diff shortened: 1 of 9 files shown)


ci/pipelines/pr.yml
```diff
@@ -1387,11 +1386,12 @@ jobs:
         outputs:
           - name: pacts
         run:
-          path: sh
+          path: bash
           dir: src
           args:
             - -ec
             - |
+              # shellcheck disable=SC1091
```
oswaldquek, line 1394: do we actually need this comment?

ci/tasks/publish-pacts.yml
```diff
@@ -24,14 +24,16 @@ run:
       version="$(cat src/.git/resource/head_sha)"
       pr="$(cat src/.git/resource/pr)"
       cd pacts || exit 1
+
+      # shellcheck disable=SC2010
```
oswaldquek, line 28: do we need this comment?

ci/pipelines/pr.yml
katstevens: Yep, this comment is what skips the shellcheck failure. Unless we can fix it instead?

oswaldquek: Nah, it's ok. I'll approve the PR.

oswaldquek (approved):
LGTM

katstevens:
This solved the shellcheck and resource failures, however the Concourse validation warnings are still causing the pipeline to fail (despite being just warnings!).

PR #371, opened 2021-02-16
PP-7683: Push nginx-forward-proxy to test ecr

oswaldquek:
New nginx-forward-proxy images are detected based on a new semantic version
(appended with the "alpine" tag) found in the
govukpay/nginx-forward-proxy repo in dockerhub. To get this to work we needed
to use a specific version of registry-image-resource as the default one got by
concourse is an older version:
```
  - name: updated-registry-image
    type: registry-image
    source:
      repository: concourse/registry-image-resource
      tag: "1.1.0"
```

We have settled with doing this because it does not have its own repository like the
other apps do. Rather, it is currently in the pay-infra project, so it would be
incorrect to tag the pay-infra project. As this nginx-foward-proxy is only used
in frontend and rarely changes, we have decided that a short term solution
would be to listen on the (semvar)-alpine tag.

The longer term solution here might be to have the nginx-forward-proxy in its
own repo and have concourse build it and tag the repo appropriately (e.g.
"alpha_release-xxx"). Then we detect nginx-forward-proxy in a similar way to
how the other apps are currently being detected.

### DIFF
--- a/ci/pipelines/apps-test-ecr.yml
+++ b/ci/pipelines/apps-test-ecr.yml
@@ -102,6 +102,14 @@ resources:
       uri: https://github.com/alphagov/pay-nginx-proxy
       branch: master
       tag_regex: "alpha_release-(.*)"
+  - name: nginx-forward-proxy-dockerhub-release
+    type: updated-registry-image
+    icon: docker
+    source:
+      variant: alpine
+      repository: govukpay/nginx-forward-proxy
+      username: ((docker-username))
+      password: ((docker-password))
 
   # ECR registry resources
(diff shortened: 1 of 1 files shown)


oswaldquek:
It's working: https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/apps-test-ecr/jobs/nginx-forward-proxy-image-to-test-ecr/builds/9

mrlumbu (approved):
Neat solution to the problem. As you'll say we'll have to come back to this, but this gets is in a good place for now

PR #376, opened 2021-02-17
PP-7794 Split toolbox test pipeline in to build and deploy

mrlumbu:
Splits existing toolbox test pipeline into two:

First pipeline triggers on git release, then takes an image from
dockerhub and pushes it to test ecr.
The only change here is that on pushing to test ecr, we add a tag
to the image of the form 123-release. This is in order to trigger
the second pipeline

The second pipeline is triggered by a push to test ecr of an image
with tag of the form `123-release`. THis pipeline has two dummy
jobs to deploy and smoke test this new image (these were previously
tasks, but it makes things clearer to have them as jobs). It then
pushes the image to staging ecr. No tags are added at this stage,
since the staging pipeline only needs a tag of form above (and this
also allows us to refer builds back to originating code change).

### DIFF
--- a/ci/pipelines/apps-test-ecr.yml
+++ b/ci/pipelines/apps-test-ecr.yml
@@ -1,4 +1,4 @@
-aws_config: &aws_config
+aws_test_config: &aws_test_config
   aws_access_key_id: ((readonly_access_key_id))
   aws_secret_access_key: ((readonly_secret_access_key))
   aws_session_token: ((readonly_session_token))
@@ -6,6 +6,14 @@ aws_config: &aws_config
   aws_role_arn: arn:aws:iam::((pay_aws_test_account_id)):role/concourse
   # Hardcode the test account registry ID for now. Needs to be a string, not a number
   aws_ecr_registry_id: "((pay_aws_test_account_id))"
+  aws_region: eu-west-1
+
+aws_staging_config: &aws_staging_config
+  aws_access_key_id: ((readonly_access_key_id))
+  aws_secret_access_key: ((readonly_secret_access_key))
+  aws_session_token: ((readonly_session_token))
+  aws_role_arn: arn:aws:iam::((pay_aws_staging_account_id)):role/concourse
+  aws_ecr_registry_id: "((pay_aws_staging_account_id))"
   aws_region: eu-west-1
 
 resources:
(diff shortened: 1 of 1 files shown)


ci/pipelines/apps-test-ecr.yml
```diff
@@ -8,6 +8,14 @@ aws_config: &aws_config
   aws_ecr_registry_id: "((pay_aws_test_account_id))"
   aws_region: eu-west-1
 
+aws_staging_config: &aws_staging_config
```
katstevens, line 11: Should we rename the other block to `aws_test_config` for consistency?

ci/pipelines/apps-test-ecr.yml
```diff
@@ -8,6 +8,14 @@ aws_config: &aws_config
...
+  aws_access_key_id: ((readonly_access_key_id))
+  aws_secret_access_key: ((readonly_secret_access_key))
+  aws_session_token: ((readonly_session_token))
+  aws_role_arn: arn:aws:iam::((pay_aws_staging_account_id)):role/concourse
+  aws_ecr_registry_id: "((pay_aws_staging_account_id))"
```
katstevens, line 16: Has this variable already been set in Concourse secrets?

ci/pipelines/apps-test-ecr.yml
```diff
@@ -117,6 +125,7 @@ resources:
     icon: docker
     source:
       repository: govukpay/toolbox
+      variant: release
```
katstevens, line 128: This is matching on the suffix, correct? `123-release` and so on?

ci/pipelines/apps-test-ecr.yml
```diff
@@ -217,6 +234,9 @@ groups:
   - name: toolbox
     jobs:
       - toolbox-image-to-test-ecr
```
katstevens, line 236: Should we rename this job for consistency too? `push-toolbox-to-test-ecr`?

ci/pipelines/apps-test-ecr.yml
```diff
@@ -217,6 +234,9 @@ groups:
...
+      - deploy-toolbox
+      - smoke-test-toolbox
+      - push-toolbox-to-staging-ecr
```
katstevens, line 239: I think this definitely looks more practical in terms of stages we'd potentially want to re-run if there was a failure 👍

katstevens (commented):
Changes look very sensible - a couple of questions.

ci/pipelines/apps-test-ecr.yml
```diff
@@ -316,30 +345,34 @@ jobs:
...
+        trigger: true
+        passed: [smoke-test-toolbox]
+      - put: toolbox-ecr-registry-staging
         params:
           image: image/image.tar
-          additional_tags: docker_tags/docker_tags.txt
+
   - name: frontend-image-to-test-ecr
     plan:
       - get: apps-test-ecr
```
danworth, line 378: Does each job need to `get: apps-test-ecr`, isn't this the git resource watching the config of this pipeline itself?

Screenshot 2021-02-17 at 16 02 31

ci/pipelines/apps-test-ecr.yml
dj-maisy: It took me a minute to suss this out as well. Why couldn't they have just called this particular key "tag_suffix" or something?

ci/pipelines/apps-test-ecr.yml
mrlumbu: yep

ci/pipelines/apps-test-ecr.yml
mrlumbu: yep - see https://github.com/concourse/registry-image-resource#source-configuration

ci/pipelines/apps-test-ecr.yml
mrlumbu: I was thinking that - but it does something slightly different. But yeah, on second thoughts I think you're right

ci/pipelines/apps-test-ecr.yml
mrlumbu: good idea, changed

ci/pipelines/apps-test-ecr.yml
katstevens: Cool 👍 We can bring the other apps in line with the naming scheme when we get to them.

katstevens (approved):
LGTM 👍

ci/pipelines/apps-test-ecr.yml
danworth: sorry, a) this PR is for toolbox b) its used in `*pull-image-from-dockerhub`...

PR #377, opened 2021-02-18
PP-7794 Parse release tags task

mrlumbu:
Generic task that takes a git release (assuming it is tagged `alpha_release-XXX`) and writes out a docker tag of form `XXX-release`. We can use input mapping to use this for all our pipelines.

### DIFF
--- a/ci/tasks/parse-release-tag.yml
+++ b/ci/tasks/parse-release-tag.yml
@@ -1,0 +1,16 @@
+platform: linux
+image_resource:
+  type: registry-image
+  source:
+    repository: alpine
+inputs:
+  - name: git-release
+outputs:
+  - name: tags
+run:
+  path: sh
+  args:
+    - -ec
+    - |
+      RELEASE_NUMBER=$(cat git-release/.git/ref | sed 's/alpha_release-//')
+      echo "${RELEASE_NUMBER}-release" > tags/tags


ci/tasks/parse-release-tag.yml
```diff
@@ -0,0 +1,16 @@
...
+  - name: git-release
+outputs:
+  - name: tags
+run:
+  path: sh
+  args:
+    - -ec
+    - |
+      RELEASE_NUMBER=$(cat git-release/.git/ref | sed 's/alpha_release-//')
+      echo "${RELEASE_NUMBER}-release" > tags
```
danworth, line 16: does this need to be `> tags/tag` ? I'm looking at this thinking that `tags` is the output directory and it needs to write into a file within that directory?

nimalank7, line 16: I think you need to `cd tags` into the directory otherwise it will create it in the current working directory. Since the tags folder is pre-created it won't let you do echo "..." > tags as the tags name is already taken (I tried to do that on my Mac with creating a tags folder and doing echo "..." which it wouldn't let me do.)

mrlumbu: yep. KInda forgot how outputs actually work

nimalank7 (approved):
LGTM!

mrlumbu:
I don't understand it's objection to `cat`

danworth:
I think its based on this https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat but opinion seems divided and sometimes reading `cat <file> | ` is easier to comprehend than file redirects or accidentally using `>` instead of `<`

PR #381, opened 2021-02-19
PP-7794 split remaining app pipelines 2

mrlumbu:
Splits products-ui, publicauth, publicapi and selfservice  pipelines as per pattern established by toolbox. Results have been applied so you can look at concourse to see if things look right


oswaldquek (approved):
LGTM

PR #385, opened 2021-02-23
PP-7800 jsify assume role task

mrlumbu:
Use node instead of python + bash to do assume role task. Makes it more maintainable, and clearer.

### DIFF
--- a/ci/scripts/assume-role.js
+++ b/ci/scripts/assume-role.js
@@ -1,0 +1,22 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const AWS = require('aws-sdk')
+
+const sts = new AWS.STS()
+
+const run = async function run() {
+  const assumeRoleResponse = await sts.assumeRole({
+    RoleArn: process.env.AWS_ROLE_ARN,
+    RoleSessionName: process.env.AWS_ROLE_SESSION_NAME
+  }).promise()
+  const tempCreds = assumeRoleResponse.Credentials
+
+  fs.writeFileSync('assume-role/assume-role.json', JSON.stringify({
+    AWS_ACCESS_KEY_ID: tempCreds.AccessKeyId,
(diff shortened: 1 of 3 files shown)


ci/scripts/assume-role.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const AWS = require("aws-sdk")
```
oswaldquek, line 4: Should probably standardise on single quotes `'` here.

ci/scripts/assume-role.js
```diff
@@ -0,0 +1,23 @@
...
+    RoleSessionName: process.env.AWS_ROLE_SESSION_NAME
+  }).promise()
+
+  const {
+    AccessKeyId: AWS_ACCESS_KEY_ID,
+    SecretAccessKey: AWS_SECRET_ACCESS_KEY,
+    SessionToken: AWS_SESSION_TOKEN
+  } = assumeRoleResponse.Credentials
+
+  fs.writeFileSync('assume-role/assume-role.yml', JSON.stringify({ AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN }))
```
oswaldquek, line 20: It's a bit confusing we're writing json to a yml file. `assume-role.json` perhaps?

oswaldquek (changes requested):
Some minor comments but the important thing is that it works 🎉

ci/scripts/assume-role.js
mrlumbu: yeah I was just copying the previous thing. I'll change it

ci/scripts/assume-role.js
mrlumbu: 👍

oswaldquek (approved):

PR #386, opened 2021-02-24
bau: Use the common assume-role.yml

oswaldquek:
Green build: https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/toolbox-staging-deploy/jobs/toolbox-staging-deploy/builds/44

### DIFF
--- a/ci/pipelines/toolbox-staging-deploy.yml
+++ b/ci/pipelines/toolbox-staging-deploy.yml
@@ -7,6 +7,12 @@ resources:
       branch: master 
       paths:
         - ci/pipelines/toolbox-staging-deploy.yml
+  - name: pay-ci
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: master
   - name: toolbox-ecr-registry-prod
     type: registry-image-resource-1-1-0
     icon: docker
@@ -52,6 +58,17 @@ jobs:
         trigger: true
         params:
           format: oci
+      - get: pay-ci
+      - task: assume-role
+        file: pay-ci/ci/tasks/assume-role.yml
+        params:
+          AWS_ROLE_ARN: arn:aws:iam::((pay_aws_staging_account_id)):role/concourse
+          AWS_ROLE_SESSION_NAME: terraform-staging-assume-role
+      - load_var: role
+        file: assume-role/assume-role.json
+        format: json
+      - load_var: digest
+        file: toolbox-ecr-registry-staging/digest
       - task: pull-tags-from-ecr-registry-staging
         config:
           platform: linux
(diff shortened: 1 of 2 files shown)


ci/pipelines/toolbox-staging-deploy.yml
```diff
@@ -52,6 +58,17 @@ jobs:
         trigger: true
         params:
           format: oci
+
+      - get: pay-ci
+      - task: assume-role
+        file: pay-ci/ci/tasks/assume-role.yml
```
mrlumbu, line 64: this needs to be 
```suggestion
        file: pay-ci/ci/tasks/assume-role.json
```

ci/pipelines/toolbox-staging-deploy.yml
```diff
@@ -78,17 +93,8 @@ jobs:
             path: /bin/sh
             args:
             - -ec
-            - | 
+            - |
               DIGEST=$(cat toolbox-ecr-registry-staging/digest)
```
mrlumbu, line 97: you could load this digest as a var as well if you want to be really nifty (see how we load the tag in deploy job)

ci/pipelines/toolbox-staging-deploy.yml
```diff
@@ -60,13 +77,11 @@ jobs:
...
-            AWS_SESSION_TOKEN: ((readonly_session_token))
-            AWS_ASSUME_ROLE_ARN: arn:aws:iam::((pay_aws_staging_account_id)):role/concourse
-            AWS_REGISTRY_ID: "((pay_aws_staging_account_id))"
+            AWS_ACCESS_KEY_ID: ((.:role.AWS_ACCESS_KEY_ID))
+            AWS_SECRET_ACCESS_KEY: ((.:role.AWS_SECRET_ACCESS_KEY))
+            AWS_SESSION_TOKEN: ((.:role.AWS_SESSION_TOKEN))
             AWS_REGION: eu-west-1
-            AWS_PAGER: ""
+            AWS_REGISTRY_ID: "((pay_aws_staging_account_id))"
             AWS_DEFAULT_OUTPUT: text
```
mrlumbu, line 85: suspect this isn't needed now

ci/pipelines/toolbox-staging-deploy.yml
```diff
@@ -78,17 +93,8 @@ jobs:
...
-              --role-arn $AWS_ASSUME_ROLE_ARN \
-              --role-session-name ecr-test-session \
-              --query 'Credentials.[AccessKeyId,SecretAccessKey,SessionToken]' \
-              ) )
-
-              export AWS_ACCESS_KEY_ID=${assumed_role[0]}
-              export AWS_SECRET_ACCESS_KEY=${assumed_role[1]}
-              export AWS_SESSION_TOKEN=${assumed_role[2]}
 
               aws configure set aws_access_key_id $AWS_ACCESS_KEY_ID --profile ecr-test-session
```
mrlumbu, line 99: these `aws configure …` lines aren't needed now

mrlumbu (changes requested):
Some small changes (mainly deleting more stuff)

ci/pipelines/toolbox-staging-deploy.yml
oswaldquek: It appears we actually need this otherwise we get the error here https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/toolbox-staging-deploy/jobs/toolbox-staging-deploy/builds/43

ci/pipelines/toolbox-staging-deploy.yml
oswaldquek: Done

ci/pipelines/toolbox-staging-deploy.yml
oswaldquek: I did delete these originally, but the build failed. I suspect the `--profile ecr-test-session` bit is key here.

mrlumbu: which build failed because of this? Can't see it

ci/pipelines/toolbox-staging-deploy.yml
mrlumbu: ah ok, yeah makes sense

ci/pipelines/toolbox-staging-deploy.yml
mrlumbu: ah so you need to remove https://github.com/alphagov/pay-ci/blob/75683f957e396eba510103704e0208e081d80bf9/ci/pipelines/toolbox-staging-deploy.yml#L103 as well

mrlumbu (approved):

PR #388, opened 2021-02-24
PP-780X Add deployment pipelines for other apps

mrlumbu:
Copy pasta of toolbox deployment - there shouldn't be any differences


danworth (approved):
Looks good 👍

PR #391, opened 2021-03-01
PP-7834: Scripts to get git sha for a release tag

oswaldquek:
In order to tag a pact with something (eg "test-fargate"), we need the version
which is the git commit sha of the release tag.

In line with the agreed practice of writing scripts in node, we have decided to
make use of a github node api. To this end we have added the `octokit/rest` npm
library to our node-aws-sdk-runner and renamed it to node-runner.

Example successful build: https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/apps-test-ecr/jobs/frontend-pact-tag/builds/3

### DIFF
--- a/ci/scripts/get-git-sha-for-release-tag.js
+++ b/ci/scripts/get-git-sha-for-release-tag.js
@@ -1,0 +1,18 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const { Octokit } = require('@octokit/rest');
+const octokit = new Octokit()
+const { TAG: imageTag, APP_NAME: appName} = process.env
+
+async function run() {
+  const gitTag = `alpha_release-${imageTag}`.replace('-release','')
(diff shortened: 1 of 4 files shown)


ci/scripts/get-git-sha-for-release-tag.js
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const { Octokit } = require('@octokit/rest');
+const octokit = new Octokit()
+
```
mrlumbu, line 6: maybe do 
```suggestion
{ TAG: tag, APP_NAME: appName} = process.env
```

ci/scripts/get-git-sha-for-release-tag.js
```diff
@@ -0,0 +1,17 @@
...
+const run = async function run() {
```
mrlumbu, line 7: after consultation with Steven Fountain esq, we decided assigning to variable here is a bit OTT, so suggest
```suggestion
async function run () {
```

mrlumbu (approved):

PR #396, opened 2021-03-02
PP-7834: Pact compatibility check for all other apps

oswaldquek:
How to review: check the relevant deploy-"app-name" jobs in https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/apps-test-ecr, verifying that the check-pact-compatibility task has run successfully.


mrlumbu (approved):
Can see the tasks for all relevant apps have been applied and are working as expected 👍

PR #398, opened 2021-03-03
PP-7686 Add deployment to staging toolbox pipeline

mrlumbu:
Add deploy step to toolbox staging pipeline. I have also split the pipeline up similarly to test pipeline, and introduced job groups. As part of this I have removed the task that passed through additional tags from the staging image through to the prod one. This is because we now don't pass any such tags from upstream. If we need to e.g. retrieve the git sha for code corresponding to an image we have a task that can do that.

### DIFF
--- a/ci/pipelines/toolbox-staging-deploy.yml
+++ b/ci/pipelines/toolbox-staging-deploy.yml
@@ -1,3 +1,20 @@
+definitions:
+  aws_assumed_role_creds: &aws_assumed_role_creds
+    AWS_ACCESS_KEY_ID: ((.:role.AWS_ACCESS_KEY_ID))
+    AWS_SECRET_ACCESS_KEY: ((.:role.AWS_SECRET_ACCESS_KEY))
+    AWS_SESSION_TOKEN: ((.:role.AWS_SESSION_TOKEN))
+
+  wait_for_deploy_params: &wait_for_deploy_params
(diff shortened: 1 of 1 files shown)


ci/pipelines/toolbox-staging-deploy.yml
```diff
@@ -43,22 +68,30 @@ resource_types:
...
         file: toolbox-staging-deploy-config/ci/pipelines/toolbox-staging-deploy.yml
-  - name: toolbox-staging-deploy
+  - name: deploy-toolbox-to-staging
     plan:
-      - get: toolbox-staging-deploy-config
       - get: toolbox-ecr-registry-staging
         trigger: true
-        params:
-          format: oci
+      - get: pay-infra
```
danworth, line 91: No need to change just a passing comment, I found myself looking where `pay-infra` is used which is within `deploy-app.yml` so its not obvious until you follow some threads or just reason that its running terraform and that's in `pay-infra` :)

danworth (approved):
Looks good to me, nice to see deploys beginning on staging 🎉

ci/pipelines/toolbox-staging-deploy.yml
mrlumbu: hmm yeah, any ways we could make this clearer? Like specify inputs in main pipeline definition or something?

PR #403, opened 2021-03-11
PP-7687 Frontend staging pipeline

mrlumbu:
Very similar to test.
No pact compatibility check as that can only be added once all apps are being pact tagged.
Removed trigger from `deploy-frontend-to-staging` job - this will be manually triggered for time being as it is in CDE (see https://docs.google.com/document/d/1oJ347kiaB_tBp-Q_7iZ1NC-xv-uQLz4etoyqI90IKVs)

### DIFF
--- a/ci/pipelines/deploy-to-staging.yml
+++ b/ci/pipelines/deploy-to-staging.yml
@@ -14,6 +14,22 @@ definitions:
     TAG: ((.:tag))
     ACCOUNT: staging
     ENVIRONMENT: staging-2
+    
+  aws_production_config: &aws_production_config
+    aws_access_key_id: ((readonly_access_key_id))
+    aws_secret_access_key: ((readonly_secret_access_key))
+    aws_session_token: ((readonly_session_token))
+    aws_role_arn: arn:aws:iam::((pay_aws_prod_account_id)):role/concourse
+    aws_ecr_registry_id: "((pay_aws_prod_account_id))"
+    aws_region: eu-west-1
+  
+  aws_staging_config: &aws_staging_config
+    aws_access_key_id: ((readonly_access_key_id))
+    aws_secret_access_key: ((readonly_secret_access_key))
+    aws_session_token: ((readonly_session_token))
+    aws_role_arn: arn:aws:iam::((pay_aws_staging_account_id)):role/concourse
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-staging.yml
```diff
@@ -14,6 +14,24 @@ definitions:
     TAG: ((.:tag))
     ACCOUNT: staging
     ENVIRONMENT: staging-2
+    
+  aws_production_config: &aws_production_config
+    aws_access_key_id: ((readonly_access_key_id))
+    aws_secret_access_key: ((readonly_secret_access_key))
+    aws_session_token: ((readonly_session_token))
+    # Need to create the role for the concourse user to assume in the production account
```
oswaldquek, line 22: do we need this comment?

ci/pipelines/deploy-to-staging.yml
```diff
@@ -14,6 +14,24 @@ definitions:
...
+    aws_role_arn: arn:aws:iam::((pay_aws_prod_account_id)):role/concourse
+    # Hardcode the production account registry ID for now. Needs to be a string, not a number
```
oswaldquek, line 24: same as above

ci/pipelines/deploy-to-staging.yml
```diff
@@ -141,4 +168,80 @@ jobs:
...
+          GIT_SHA: ((.:git-sha))
+          APP_NAME: frontend
+          PACT_TAG: staging-fargate
+  - name: push-frontend-to-production-ecr
+    plan:
+      - get: frontend-ecr-registry-staging
+        params:
+          format: oci
+        trigger: true
+        passed: [smoke-test-frontend-on-staging]
```
oswaldquek, line 243: I don't have an issue with this but probably best to make it `[frontend-pact-tag]` to keep it consistent with what we do in test.

oswaldquek (changes requested):

ci/pipelines/deploy-to-staging.yml
mrlumbu: 👍

ci/pipelines/deploy-to-staging.yml
mrlumbu: nope, removed

oswaldquek (approved):

PR #407, opened 2021-03-12
PP-7688 add more staging deployments

mrlumbu:
Adds deployment pipelines for adminusers, products and products-ui (with pact tagging) and cardid and publicauth. Follows exact same pattern established by toolbox and frontend


oswaldquek (approved):
LGTM

PR #409, opened 2021-03-15
PP-7877 - Add aws_test_config to pay deploy

nimalank7:
Description:
- For pay deploy to pull the telefrag image from test ECR it needs permissions


mrlumbu:
turns out we won't need this

PR #414, opened 2021-03-17
PP-7877 Trigger toolbox deploy on new telegraf

mrlumbu:
We plan to use Toolbox deployments as a way of ensuring new versions of sidecars are not totally broken. This commit enbables this by triggering a new deploy of toolbox when a new version of telegraf appears in test ECR. The job then deploys this new version (alongside version of toolbox specified by toolbox ecr resource) I have split into an inline task so as not to have any confusing logic in the deploy-app task.

See https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-toolbox/builds/149

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -57,7 +57,7 @@ resources:
       branch: main
       username: alphagov-pay-ci
       password: ((github-access-token))
-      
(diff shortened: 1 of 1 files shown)


danworth (approved):
nice 👍

PR #416, opened 2021-03-17
PP-7877: Push telegraf image to staging ecr

oswaldquek:
This happens upon a successful smoke test in toolbox.

with: @nimalank7 @mrlumbu


nimalank7 (approved):
LGTM!

PR #417, opened 2021-03-17
PP-7887 Trigger smoke tests for toolbox post-deploy

katstevens:
- Adds a generic task for running smoke tests, given a SMOKE_TEST_NAME corresponding to the name of the canary (this is likely to change in the near future)
- Adds the NodeJS script to trigger the canary via the AWS SDK (lifted entirely from @danworth and @rorymalcolm's [spike branch](https://github.com/alphagov/pay-ci/compare/PP-7670_spike_aws_canary), with a little tweaking for linting and error handling)
- Replaces the dummy smoke test steps with the steps to assume the trigger role for the deploy account, and call the NodeJS script for each scenario with the correct variables. 
 
Note that not all the Canaries are available yet in the AWS deploy account, only `card_payment_test` and `card_payment_staging`.

The new Concourse pipeline variable `pay_aws_deploy_account_id` has been added to the pay-dev and pay-deploy teams already.

**NOTE**: Not ready for merging yet - waiting for the Canaries to be available

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -23,6 +23,59 @@ definitions:
     aws_ecr_registry_id: "((pay_aws_prod_account_id))"
     aws_region: eu-west-1
 
+  # Separate tasks for each combination of scenario/environment
+  smoke-test-run-all-on-production: &smoke-test-run-all-on-production
+    limit: 4
+    steps:
+      - task: run_create_card_payment_sandbox-production
+        file: pay-ci/ci/tasks/run-smoke-test.yml
+        params:
+          <<: *aws_assumed_role_creds
+          AWS_REGION: "eu-west-1"
+          SMOKE_TEST_NAME: "card_sandbox_prod"
+      - task: run_create_card_payment_smartpay-production
+        file: pay-ci/ci/tasks/run-smoke-test.yml
+        params:
+          <<: *aws_assumed_role_creds
+          AWS_REGION: "eu-west-1"
+          SMOKE_TEST_NAME: "card_smartpay_prod"
+      - task: run_create_card_payment_worldpay_with_3ds-production
+        file: pay-ci/ci/tasks/run-smoke-test.yml
+        params:
+          <<: *aws_assumed_role_creds
+          AWS_REGION: "eu-west-1"
+          SMOKE_TEST_NAME: "card_wpay_3ds_prod"
+      - task: run_create_card_payment_worldpay_with_3ds2-production
+        file: pay-ci/ci/tasks/run-smoke-test.yml
+        params:
+          <<: *aws_assumed_role_creds
+          AWS_REGION: "eu-west-1"
+          SMOKE_TEST_NAME: "card_wpay_3ds2_prod"
+      - task: run_create_card_payment_worldpay_with_3ds2_exemption-production
+        file: pay-ci/ci/tasks/run-smoke-test.yml
+        params:
+          <<: *aws_assumed_role_creds
+          AWS_REGION: "eu-west-1"
+          SMOKE_TEST_NAME: "card_wpay_3ds2ex_prod"
+      - task: run_create_card_payment_worldpay_without_3ds-production
+        file: pay-ci/ci/tasks/run-smoke-test.yml
+        params:
+          <<: *aws_assumed_role_creds
+          AWS_REGION: "eu-west-1"
+          SMOKE_TEST_NAME: "card_wpay_prod"
+      - task: run_cancel_card_payment_sandbox-production
+        file: pay-ci/ci/tasks/run-smoke-test.yml
+        params:
+          <<: *aws_assumed_role_creds
+          AWS_REGION: "eu-west-1"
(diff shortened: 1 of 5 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -23,6 +23,64 @@ definitions:
     aws_ecr_registry_id: "((pay_aws_prod_account_id))"
     aws_region: eu-west-1
 
+  # Separate tasks for each combination of scenario/environment
+  smoke-test-create-card-payment-sandbox-on-production: &smoke-test-create-card-payment-sandbox-on-production
```
danworth, line 27: I'm guilty of anchoring the original pr pipeline to within (or more likely beyond) an inch of its life and in hindsight it wasn't a fabulous idea (easier to write and modify but less easy to read 6 months later). I'm just wondering if these many anchors could be replaced with just one `&run-every-smoke-test` anchor since we seem to reference them all every time (unless I've miss-read it)?

katstevens: I tried doing that and the Concourse validation complained at me for having multiple tasks in the same anchor - very happy to squish them down if there's a way to do that?

danworth: Maybe concourse validation is trying to protect us from ourselves. Was it valid yaml but the concourse `validate-pipeline` didn't like it?

katstevens: Yep. If I did something like:

```
smoke-test-run-all-on-production: &smoke-test-run-all-on-production
    - task: run_create_card_payment_sandbox-production
      file: pay-ci/ci/tasks/run-smoke-test.yml
      params:
        <<: *aws_assumed_role_creds
        AWS_REGION: "eu-west-1"
        SMOKE_TEST_NAME: "card_sandbox_prod"
    - task: run_create_card_payment_smartpay-production
      file: pay-ci/ci/tasks/run-smoke-test.yml
      params:
        <<: *aws_assumed_role_creds
        AWS_REGION: "eu-west-1"
        SMOKE_TEST_NAME: "card_smartpay_prod"
     [etc]
```

Then I'd get the following when trying to commit:

```
error: yaml: map merge requires map or sequence of maps as the value
```

mrlumbu: you could potentially solve this problem and speed everything up by running them `in_parallel`  all https://concourse-ci.org/jobs.html#in-parallel-step. That would allow you to have one anchor for the top level in parallel bit, which would contain all these bad boys

katstevens: Ahhh that's perfect. Thanks @mrlumbu!

katstevens:
I've followed @mrlumbu's suggestion and grouped the tests under a single anchor, run `in_parallel` (which seems to validate OK). I've set the `limit: 4` for now as I don't want to overwhelm the Concourse workers.

ci/scripts/run_smoke_test.js
```diff
@@ -0,0 +1,76 @@
...
+      // Check failure details from S3 bucket
+      try {
+        const s3Objects = await getS3Objects(splitLocation, prefix)
+        const logFile = s3Objects.Contents.filter(obj => obj.Key.includes('.txt')).map(obj => obj.Key)[0]
+        const logStream = s3.getObject({ Bucket: splitLocation[0], Key: logFile }).createReadStream()
+        logStream.on('readable', () => {
+          console.log(`${logStream.read()}`)
+        })
+      } catch (error) {
+        console.log(error)
```
danworth, line 66: This could be something like
```
console.error("Failed to get details from S3:", error.message)
```

ci/scripts/run_smoke_test.js
```diff
@@ -0,0 +1,76 @@
...
+async function getS3Objects (splitLocation, prefix) {
+  return s3.listObjects({ Bucket: splitLocation[0], Prefix: prefix }).promise()
+}
+
+async function run () {
+  const startedAt = Date.now()
+  try {
+    await runCanary()
+  } catch (error) {
+    console.error(error)
```
danworth, line 36: Could add some context to the error
```
console.error("Failed to run canary: ", error.message)
```

ci/scripts/run_smoke_test.js
```diff
@@ -0,0 +1,76 @@
...
+    process.exitCode = 1
```
danworth, line 37: Not sure which is best practice (interested to know), but I've used the following before rather than `process.exitCode = 1` and `return`
```
process.exit(1)
```

danworth (approved):
Looks good to me. There's quite a bit in there but I'm sure we'll give it a good check once its up and running to make sure it behaves as expected. nice yaml :)

🏅 🏅

danworth:
> I've followed @mrlumbu's suggestion and grouped the tests under a single anchor, run `in_parallel` (which seems to validate OK). I've set the `limit: 4` for now as I don't want to overwhelm the Concourse workers.

Yeah I think the nice thing about how we're doing this, is that AWS/canaries are doing all the hard work and the node running on Concourse isn't doing much. Good to be cautious and see how it goes 👍

PR #419, opened 2021-03-17
PP-7877 - Pass telegraf version to Terraform to deploy Toolbox to sta…

nimalank7:
…ging

Description:
- In order to have Toolbox deployed with a new version of telegraf we need to pass the telegraf version number into the Terraform from the tag in telegraf test ECR
- The task is now inline rather than using a file


oswaldquek (approved):
Looks like https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-staging/jobs/deploy-toolbox-to-staging/builds/22 passed the deploy stage so LGTM.

PR #426, opened 2021-03-17
PP-7877 - Show deploy-toolbox and deploy-toolbox-to-staging in telegraf group

nimalank7:
Description:
- Since the telegraf image will be deployed to both staging and production ECR conditional on passing the deploy-toolbox
and deploy-toolbox-staging jobs it makes sense to display it in the UI for clarity

### DIFF
--- a/ci/pipelines/deploy-to-staging.yml
+++ b/ci/pipelines/deploy-to-staging.yml
@@ -287,6 +287,7 @@ groups:
       - push-ledger-to-production-ecr
   - name: telegraf
     jobs:
+      - deploy-toolbox-to-staging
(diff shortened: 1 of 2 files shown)


mrlumbu:
we could do same for test as well?

nimalank7:
Do you mean building the docker image and pushing to test? That isn't conditional on anything in toolbox passing. I've added the jobs to both the deploy-to-test and deploy-to-staging which are the only cases where we wait on Toolbox deployment

mrlumbu:
sorry, totally didn't read properly

mrlumbu (approved):

PR #427, opened 2021-03-18
Pp 7688 add pact compatibility checks

mrlumbu:
Now we have pact tagging in all pipelines we can do pact compatibility checks (as we already do in test pipeline) to ensure compatibility of services in a given environment.

See https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-staging/jobs/deploy-adminusers-to-staging/builds/2 e.g. for example of this working


rorymalcolm (approved):
this LGTM

PR #428, opened 2021-03-18
PP-7877 - Fix authentication error with production ECR

nimalank7:
Description:
- Some production ECR resources were missing the role configuration which caused an authentication error when checking.
- Toolbox should only check for `*-release` Telegraf variants as that is what the docker images are tagged with

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -114,6 +114,8 @@ resources:
     icon: docker
     source:
       repository: govukpay/telegraf
+      variant: release
+      <<: *aws_prod_config
   - name: ledger-ecr-registry-prod
     type: registry-image-resource-1-1-0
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -126,7 +128,7 @@ resources:
     icon: docker
     source:
       repository: govukpay/publicapi
-
+      <<: *aws_prod_config
```
mrlumbu, line 131: also needs `variant: release` I think

mrlumbu (changes requested):
One change needed

ci/pipelines/deploy-to-production.yml
nimalank7: Fixed

mrlumbu (approved):

PR #430, opened 2021-03-18
PP-7702 enable my services as default view for selfservice

SandorArpa:
- enable default view


danworth (approved):
Looks good, this is true for selfservice in all environments 🏅

PR #436, opened 2021-03-18
PP-7755 Test new versions of nginx with toolbox deploy

mrlumbu:
As with other sidecars, on new version of nginx proxy then deploythe new version as by passing in new version to toolbox deploy, and only push the new image to staging once that deploy has succeeded.

https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test?group=nginx-proxy


nimalank7 (approved):
LGTM!

PR #438, opened 2021-03-18
PP-7755 nginx check staging prod

mrlumbu:
As with telegraf, deploy new version of nginx in staging as part of toolbox deployment to check its ok before pushing up to prod, then deploy to toolbox in prod to check its ok in prod.

https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-staging?group=nginx-proxy
https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-production?group=nginx-proxy

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -116,6 +116,13 @@ resources:
       repository: govukpay/telegraf
       variant: release
       <<: *aws_prod_config
+  - name: nginx-proxy-ecr-registry-prod
+    type: registry-image-resource-1-1-0
+    icon: docker
+    source:
+      repository: govukpay/docker-nginx-proxy
+      variant: release
+      <<: *aws_prod_config
   - name: ledger-ecr-registry-prod
     type: registry-image-resource-1-1-0
     icon: docker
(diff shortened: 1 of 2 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -194,6 +201,9 @@ groups:
       - deploy-ledger-to-prod
       - smoke-test-ledger-on-prod
       - ledger-pact-tag
+  - name: nginx-proxy
+    jobs:
+      - deploy-toolbox-to-prod
```
nimalank7, line 207: I don't think we need this. We need it for deploy-to-staging and deploy-to-test because the push-to-staging/prod is conditional on the deploy-toolbox-to-staging/prod. We could always just check toolbox to see if the deploy to prod has been successful.

mrlumbu: I think it makes it clear that when a new version of nginx is pushed, this job is triggered, rather than having to know that its hiding in the toolbox job group

nimalank7: Sure. To be consistent then we should do this for the nginx-forward-proxy and telegraf

nimalank7 (approved):
LGTM!

PR #439, opened 2021-03-18
PP-7755 - Trigger Frontend deployment to test on new nginx-forward-proxy image in test ECR

nimalank7:
Description:
- Adds a trigger to frontend to deploy to test when a new nginx-forward-proxy image appears in test ECR

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -254,6 +254,7 @@ resources:
     icon: docker
     source:
       repository: govukpay/nginx-forward-proxy
+      variant: release
       <<: *aws_test_config
   - name: telegraf-ecr-registry-test
     type: dev-registry-image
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -1919,6 +1952,18 @@ jobs:
...
           image: image/image.tar
           additional_tags: nginx-forward-proxy-git-release/.git/ref
+  - name: push-nginx-forward-proxy-to-staging-ecr
+    plan:
+      - get: pay-ci
+      - get: nginx-forward-proxy-ecr-registry-test
+        params:
+          format: oci
+        trigger: true
+        passed: [deploy-frontend]
```
mrlumbu, line 1962: this could actually be smoke test frontend, since we do have smoke tests for frontend

nimalank7: Good shout yes will change that.

mrlumbu (approved):
Nice

PR #445, opened 2021-03-19
PP-7889 Notify success and failure of toolbox test deployments

mrlumbu:
Couple of funny looking things:
The message is partly comprised of a bit written to a file and then loaded to a variable. This allows us to move the task that creates the messages to a task file, parameterise etc.
The other bit of the message (the build information) is stuck on the end in the `on_{success,failure}` handlers. This is because Concourse only exposes build information in resources https://stackoverflow.com/a/40939590 so if this were appended to the message in the create messages task, all the env vars would evaluate to an empty string.
Success goes to `govuk-pay-activity`, failure to `govuk-pay-announce` as discussed in story notes.

Example of success being sent https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-toolbox/builds/169
Example of success message: https://gds.slack.com/archives/CEBPUR2JE/p1616166796010200

Example of failure being sent https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-toolbox/builds/170

Example failure message: https://gds.slack.com/archives/CEBPUR2JE/p1616167252010800 (nb have fixed the :red_circle: emoji)

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -348,6 +348,12 @@ resources:
       repository: govukpay/nginx-forward-proxy
       variant: release
       <<: *aws_staging_config
+  - name: slack-notification
+    type: slack-notification
+    source:
+      url: https://hooks.slack.com/services/((slack-notification-secret))
+
+
 
 resource_types:
   # Custom resource type - this has been merged to the
@@ -367,7 +373,11 @@ resource_types:
     source:
       repository: concourse/registry-image-resource
       tag: "1.1.0"
-
+  - name: slack-notification
+    type: docker-image
+    source:
+      repository: cfcommunity/slack-notification-resource
+      tag: latest
 groups:
   - name: update-deploy-to-test-pipeline
     jobs:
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -526,6 +539,45 @@ jobs:
...
+            TELEGRAF_GIT_TAG: ((.:telegraf_image_tag)
+          outputs:
+            - name: message
+          run:
+            path: sh
+            args:
+              - -c
+              - |
+                cat <<EOT >> message/success
+                :green-circle: Deployment of ${APP_NAME} was successful :tada:
```
danworth, line 563: I suppose this could mention the environment as well (same for the error scenario)?

danworth (approved):
Looks good 👍 

Wonder if we can mention the environment and thinking about it whether we should add something about these apps are not live yet to don't panic if you see a failed build for people not in our team. I guess until we get closer to going live the failure notifications are more for our team's benefit.

ci/pipelines/deploy-to-test.yml
mrlumbu: yes you're right it should mention the environment 😬

mrlumbu:
Yeah I suppose we could have a top level disclaimer that this is FAO team starling for the moment

mrlumbu:
I'm gonna merge this and fix those things as I apply to everything else if that's ok

PR #446, opened 2021-03-19
PP-7915 - Frontend deploys with newest nginx version in test

nimalank7:
Description:
- Frontend now deploys with the latest nginx proxy in test
- Refactors the inline task into a separate file called deploy-frontend.yml


oswaldquek (approved):
LGTM

PR #447, opened 2021-03-22
PP-7915 - Latest version of nginx for all test apps

nimalank7:
Description:
- Remaining apps will be deployed to test with the latest version of nginx proxy


nimalank7:
This needs to be done for staging and prod otherwise the pipeline will break.

PR #449, opened 2021-03-22
PP-7915 - Latest version of nginx for all test apps

nimalank7:
Description:
- Remaining apps will be deployed to test with the latest version of nginx proxy

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -7,11 +7,13 @@ definitions:
   wait_for_deploy_params: &wait_for_deploy_params
     <<: *aws_assumed_role_creds
     TAG: ((.:tag))
+    NGINX_IMAGE_TAG: ((.:nginx_image_tag))
     ENVIRONMENT: production-2
 
   deploy_params: &deploy_params
     <<: *aws_assumed_role_creds
     TAG: ((.:tag))
+    NGINX_IMAGE_TAG: ((.:nginx_image_tag))
     ACCOUNT: production
     ENVIRONMENT: production-2
 
@@ -37,7 +39,7 @@ resources:
     icon: github
     source:
       uri: https://github.com/alphagov/pay-ci
-      branch: master
+      branch: PP-7915-Latest-nginx-all-env
   - name: pay-infra
     type: git
     icon: github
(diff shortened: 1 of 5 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -12,6 +12,7 @@ definitions:
   deploy_params: &deploy_params
     <<: *aws_assumed_role_creds
     TAG: ((.:tag))
+    NGINX_IMAGE_TAG: ((.:nginx_image_tag))
```
katstevens, line 15: I do wonder if we should rename `TAG` to `APPLICATION_IMAGE_TAG` (not in this PR though!)

nimalank7: We definitely should! I didn't do it before because would break some stuff but I'll be able to do it after this story is finished.

katstevens: Do we want to include `NGINX_IMAGE_TAG` in the `wait_for_deploy_params` as well?

ci/pipelines/deploy-to-production.yml
```diff
@@ -499,7 +506,6 @@ jobs:
         file: pay-ci/ci/tasks/deploy-frontend.yml
         params:
           APP_NAME: frontend
-          NGINX_IMAGE_TAG: ((.:nginx_image_tag))
```
katstevens, line 502: Do we want to use the `<<: *deploy_params` anchor for toolbox as well? I think we're still writing them out long form there?

ci/pipelines/deploy-to-staging.yml
```diff
@@ -1174,11 +1198,13 @@ jobs:
   - name: deploy-ledger-to-staging
     plan:
       - get: ledger-ecr-registry-staging
-        trigger: true
```
katstevens, line 1177: How come we're removing the trigger here?

ci/pipelines/deploy-to-production.yml
nimalank7: Good spot! Yes - all the wait-for-deploy.js script is doing is just printing out the value of `TAG` so probably worth having the version of nginx alongside with it as well. Frontend is unique as it has an additional nginx proxy so might have to modify wait-for-deploy.js to display the 2nd nginx as well...

ci/pipelines/deploy-to-production.yml
nimalank7: Yes a follow up PR will do this because Toolbox is also deploying telegraf which `<<: *deploy_params` doesn't have at the moment.

ci/pipelines/deploy-to-staging.yml
nimalank7: Ah it shouldn't be in production either! We should only be triggering things into staging + prod which aren't in the CDE.

ci/pipelines/deploy-to-production.yml
katstevens: Ah that makes sense 👍

katstevens (approved):
Great stuff 👍

PR #453, opened 2021-03-22
PP-7887: Add pay-ci resource to smoke tests

katstevens:
Fixes an oversight - this was included in the `deploy-to-test` pipeline but not the equivalent staging or production pipelines.


nimalank7 (approved):
LGTM!

PR #454, opened 2021-03-23
PP-7915 - Display nginx versions in deploy script

nimalank7:
Description:
- Display the nginx proxy version as output when the wait-for-deploy.js script runs
- If the app is frontend then also display the nginx frontend version as well otherwise don't


oswaldquek (approved):
nice one!

PR #455, opened 2021-03-23
PP-7915 - Latest telegraf for all apps

nimalank7:
Description:
- All apps now deploy using the latest version of telegraf which is passed into the Terraform

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -26,12 +26,14 @@ wait_for_deploy_params: &wait_for_deploy_params
   <<: *aws_assumed_role_creds
   TAG: ((.:tag))
   NGINX_IMAGE_TAG: ((.:nginx_image_tag))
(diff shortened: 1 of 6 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -587,7 +589,7 @@ jobs:
         trigger: true
       - get: pay-infra
       - get: pay-ci
-      - load_var: toolbox_image_tag
+      - load_var: tag
```
mrlumbu, line 592: Is there a reason you've gone back to calling this `tag` rather than say `app_image_tag`?

mrlumbu (approved):
Looks good. One small comment that could be addressed in a follow up PR

ci/pipelines/deploy-to-test.yml
nimalank7: Yep originally the deploy for Toolbox was an inline tag. Since I'm using the `deploy_params` and `deploy-app.yml` has `TAG` I had to change it to that. A future PR will rename `tag` to `app_image_tag`.

PR #458, opened 2021-03-24
PP-7887: Check Canary is in a startable state

katstevens:
When running our post-deploy smoke tests, if second app is also being deployed, and the smoke tests are already running, the `startCanary` command will fail with a 409 Conflict error.

Ideally we want the Canary to be in status STOPPED, so we can be confident that our forthcoming run will test the just-deployed app. This script adds a check on the Canary's current status, and happily proceeds if the status is STOPPED.

If the status is DELETING or ERROR, something bad has happened to our Canary and we probably shouldn't run it - raise an error.

If the status is anything else (eg RUNNING, STARTED, UPDATING) then waiting for a minute or so should bring the Canary into the STOPPED state. Our Canaries currently take way less than 60s to run and/or update, so this is fairly generous and should cover most cases. After waiting, we can call `startCanary` as normal.

Here's what this scenario looks like in Concourse (note the 60s gap in the timestamp):

concourse-smoke-test-wait

If the `StartCanary` command still fails with a conflict after 60s, then it's likely to be conflicting with a fresh run (maybe triggered by a third app deploy!). However this run will have started *after* our deploy, so it's fine to use the result. If the canary fails for any other reason, the script errors.

There's an edge case where the same canary is still running after 60s, and its results are based on the previous deploy. Given long-running canaries should time out and fail anyway, this doesn't feel like a big risk. We'll address this case and improve the error handling in a future story.

[AWS error reference for StartCanary](https://docs.aws.amazon.com/AmazonSynthetics/latest/APIReference/API_StartCanary.html)

### DIFF
--- a/ci/scripts/run_smoke_test.js
+++ b/ci/scripts/run_smoke_test.js
@@ -67,8 +67,6 @@ async function run () {
     process.exit(1)
   }
 
-  const startedAt = Date.now()
-
   if (canaryStatus === 'wait') {
     // Wait 60 secs and continue
     console.log('Canary is not in a startable state. Wait 60s before starting.')
@@ -78,6 +76,8 @@ async function run () {
       console.log(error)
     }
   }
+
+  const startedAt = Date.now()
 
   try {
     console.log('Starting Canary')
--- a/ci/scripts/run_smoke_test.js
+++ b/ci/scripts/run_smoke_test.js
@@ -4,13 +4,18 @@ const AWS = require('aws-sdk')
 const synthetics = new AWS.Synthetics()
 const s3 = new AWS.S3()
 const CHECK_INTERVAL = 5000
+const CANARY_TIMEOUT = 60000
 const { SMOKE_TEST_NAME } = process.env
 
 async function describeCanariesLastRun () {
   return synthetics.describeCanariesLastRun({}).promise()
 }
 
-function runCanary () {
+async function getCanary () {
+  return synthetics.getCanary({ Name: SMOKE_TEST_NAME }).promise()
+}
+
+function startCanary () {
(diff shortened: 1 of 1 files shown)


oswaldquek (approved):
LGTM

ci/scripts/run_smoke_test.js
```diff
@@ -28,14 +33,65 @@ async function getS3Objects (splitLocation, prefix) {
...
+  }
+
+  if (canaryStatus === 'wait') {
+    // Wait 60 secs and continue
+    console.log('Canary is not in a startable state. Wait 60s before starting.')
+    try {
+      await sleep(CANARY_TIMEOUT)
+    } catch (error) {
+      console.log(error)
+    }
```
oswaldquek, line 77: not really sure there'll be an error here but not a biggie

ci/scripts/run_smoke_test.js
```diff
@@ -28,14 +33,65 @@ async function getS3Objects (splitLocation, prefix) {
...
+      return 'ready'
+    case 'DELETING':
+    case 'ERROR':
+      throw new Error(`Unable to start canary in state ${canaryStatus}`)
+  }
+}
+
+function sleep (ms) {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
```
oswaldquek, line 56: could do `await new Promise(r => setTimeout(r, 2000))` inline I think

ci/scripts/run_smoke_test.js
katstevens: Yeah, it's probably overkill but we had a few warnings pop up about uncaught promises, so wanted to make sure it wasn't this one.

PR #460, opened 2021-03-24
PP-7889 Notifications on test pipeline for deploys, smoke tests, and pact tagging

mrlumbu:
Add success and failure notifications to all deployments, smoke tests and pact tag jobs on test pipeline.
I have extracted a task to create success and failure snippets that I think is reasonably readable and self explanatory.

For deployment notifications we can tell people about versions of all the things deployed which is probably useful.

For smoke tests and pact tags we only have the version of the app (no sidecar versions) so we just include that in notification.

At present this will send all successes to `govuk-pay-activity` and all failures to `govuk-pay-announce`. I have caveated all notifications as only FAO migration team, but I think good to start sending failures to a visible channel, both for migration team to be aware, and to start to get wider team used to this whole thing.

Example of success notification: https://gds.slack.com/archives/CEBPUR2JE/p1616588487009600
from this build https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-connector/builds/39

Screenshot 2021-03-24 at 12 34 08

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -53,6 +53,16 @@ deploy_params: &deploy_params
   ACCOUNT: test
   ENVIRONMENT: test-12
 
+snippet_params_all_versions: &snippet_params_all_versions
+  ENV: test-12
+  APP_IMAGE_TAG: ((.:tag))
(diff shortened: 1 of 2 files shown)


danworth (approved):
This is brilliant, very well thought through and the messages posted are clear 🏅

PR #462, opened 2021-03-24
PP-7889 Add notifications to staging pipeline

mrlumbu:
Following pattern on test pipeline.

Only difference is all deployment activity (including beginning of a deployment) is posted to announce channel, as per kick off notes and current dev practice

https://gds.slack.com/archives/CAEN2AVLP/p1616598908006300


danworth (approved):
🏅 looks good

PR #464, opened 2021-03-24
PP-7908 Check for downgrade

mrlumbu:
In our current Jenkins pipelines we have a check that
stops devs accidentally deploying an old version of an
app. This `check-release-versions` task achieves same thing
on Concourse - it checks the version to be deployed of each
container in a task definition, and makes sure it is the
same as or newer than currently deployed version.

There is no mechanism to override this at the moment -
if override is needed devs will need to deploy the thing
manually from laptop.

https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-toolbox/builds/209

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -636,6 +636,29 @@ jobs:
       - load_var: role
         file: assume-role/assume-role.json
         format: json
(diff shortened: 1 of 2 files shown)


oswaldquek (approved):
LGTM

PR #468, opened 2021-03-25
PP-7908 check release versions on all deployments

mrlumbu:
Extract existing check-release-versions task to, you guessed it, check release versions for all deployments in test, staging and prod pipeline.


danworth (approved):
Looks sensible 👍

PR #475, opened 2021-03-26
Fix PR pact things

mrlumbu:
For provider pact tests to run on PRs they need to be able
to pull down master/main versions of consumer contracts.
However since some time in autumn last year we have not been
tagging contracts with main/master on concourse pact broker
(for reasons to do with abandonment of PaaS migration).

This PR moves all provider pact testing gubbins to point at
the 'old' pact broker, and uses master tag (as that is what
we use in old pact broker world).

This should work...


danworth (approved):
🤞 should work

PR #477, opened 2021-03-26
BAU increase parallelism of smoke tests

mrlumbu:
Concourse is not doing the heavy lifting, so let's run 8 in parallel and see what happens


rorymalcolm (approved):
LGTM 🚀

PR #478, opened 2021-03-30
BAU: Condense slack notifications

katstevens:
- Made the Fargate element less shouty (but still a bit shouty), and linked the text rather than printing the URLs in full
- Removed the sidecar version info for 'start' and 'success' builds. People can always click on the build if they want to see the telegraf/nginx versions etc. 
- Converted full URLs to linked descriptive text where possible

These changes have been applied to the test, staging and production deploy pipelines - see `#govuk-pay-announce` and `#govuk-pay-activity` for examples.

Not done, but that we might want to consider in future:
- different green emojis for the different types of success announcement (deploys, smoke tests, pact tests)

### DIFF
--- a/ci/tasks/create-notification-snippets.yml
+++ b/ci/tasks/create-notification-snippets.yml
@@ -36,12 +36,12 @@ run:
       [ -n "${NGINX_IMAGE_TAG}" ] && \
       NGINX_RELEASE_NUMBER=$(echo ${NGINX_IMAGE_TAG} | sed 's/-release//') &&\
       echo "<https://github.com/alphagov/pay-nginx-proxy/releases/tag/alpha_release-${NGINX_RELEASE_NUMBER}|nginx-proxy v${NGINX_RELEASE_NUMBER}>" | \
-      tee -a snippet/* || true
+      tee -a snippet/failure || true
 
       [ -n "${TELEGRAF_IMAGE_TAG}" ] && \
       echo " | <https://github.com/alphagov/pay-telegraf/releases/tag/${TELEGRAF_IMAGE_TAG}|telegraf v${TELEGRAF_IMAGE_TAG}>" | \
-      tee -a snippet/* || true
+      tee -a snippet/failure || true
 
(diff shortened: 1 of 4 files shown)


ci/tasks/create-notification-snippets.yml
```diff
@@ -19,43 +19,29 @@ run:
...
     - -c
     - |
-      cat <<EOT >> snippet/start
-      THIS MESSAGE RELATES TO FARGATE - IGNORE UNLESS WORKING ON MIGRATION
-      :rocket: ${ACTION_NAME} of ${APP_NAME} on ${ENV} is beginning
+      APP_RELEASE_NUMBER=$(echo ${APPLICATION_IMAGE_TAG} | sed 's/-release//')
 
-      Version details:
+      cat <<EOT >> snippet/start
+      :rocket: FARGATE ${ACTION_NAME} of <https://github.com/alphagov/pay-${APP_NAME}/releases/tag/alpha_release-${APP_RELEASE_NUMBER}|${APP_NAME} ${APP_RELEASE_NUMBER}-release> on ${ENV} is beginning
```
danworth, line 25: Really nice changes. Maybe something for future thought, I think it might be more useful to link the commit that the release was built from than the release itself since I feel the first question I'm asking if I'm going to look at what's been deployed is, what was the change? e.g. 
https://github.com/alphagov/pay-publicapi/commit/959141185bbd35aa71013cb182f64584b2ca6b98
rather than 
https://github.com/alphagov/pay-publicapi/releases/tag/alpha_release-802

We might not have that information readily available within the job and it's simple to get to the commit from the release page so no biggie.

danworth (approved):
🏅 nice changes 👍

ci/tasks/create-notification-snippets.yml
katstevens: Agreed, that would be useful. I think you're right though, that the commit might not be easily retrievable everywhere yet. I'll add that to the Future Ideas list (shhh, don't call it tech debt... 😉 )

PR #480, opened 2021-04-28
PP-8011 Deploy apps, one release at a time

katstevens:
Previously, triggering a new release (e.g `release-123`) in Concourse before a previous release (`release-122`) had finished caused the newer release to fail (as the newer job can't acquire the terraform lock). We don't want more than 1 release at once anyway, so adding `serial: true` ensures releases happen one after the other.

I've tested this on `deploy-to-test` pipeline already and it works: newer releases are shown as pending (grey) with the message `'checking max-in-flight is not reached'` until the previous release has finished, after which they continue as normal. 

Once this PR is approved and merged, I'll apply to the `deploy-to-staging` and `deploy-to-production` pipelines.


danworth (approved):
Looks good and its been tested 👍

PR #481, opened 2021-04-28
PP-8011 run smoke test in serial groups

katstevens:
Previously, the same smoke test jobs (Canaries) could be triggered by the release of different apps. Although we have a workaround in place to retry Canaries that are already running, this can still cause some issues (and is slow, as the workaround involves waiting 60 seconds).

Instead, we can add a `serial_groups` tag for the smoke test jobs. Any job with the tag '`smoke-test`' in its `serial_groups` list will not run if another job with the same tag is in progress (see [example in the Concourse docs](https://concourse-ci.org/jobs.html#schema.job.serial_groups)).

As per https://github.com/alphagov/pay-ci/pull/480 I've already tested this on the `deploy-to-test` pipeline: I started the smoke test job for frontend, then also triggered the smoke test job for toolbox. The toolbox smoke tests waited patiently until the frontend smoke tests had finished, then started up as usual.

Merging this PR will update the pipelines for staging and production.

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -408,6 +408,7 @@ jobs:
     <<: *put_failure_slack_notification
 
   - name: smoke-test-selfservice-on-prod
+    serial_groups: [smoke-test]
     plan:
       - get: selfservice-ecr-registry-prod
         trigger: true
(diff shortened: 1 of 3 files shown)


rorymalcolm (approved):
good change 🎉

PR #490, opened 2021-05-25
PP-7967 Upgrade maven version from 3.6.1 to 3.8.1

kbottla:
## WHAT
- Upgrades to latest maven version


danworth (approved):
Seems sensible, would be good to keep an eye on a few builds after applying just to make sure, unless you've verified this via other means.

PR #497, opened 2021-06-02
PP-7922 Slack notifications for DB migration jobs

katstevens:
Pay developers have kindly requested that we add automatic Slack notifications when a DB migration job is triggered, to save them having to post in Slack manually.

- Concourse can't determine the result of a DB migration, so we don't need success/failure snippets like we have for the release jobs. This means we can hardcode a single snippet and don't need to use the `create-notification-snippet.yml` bash script.
- I've set test, staging and production to all put notifications in `#govuk-pay-announce` (as opposed to `#govuk-pay-activity`) as DB migrations are important to know about but rare enough to not cause too much noise. 
- However [I did test this change on the new publicauth DB migration job](https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/publicauth-db-migration/builds/2), temporarily going to `#govuk-pay-activity` (to avoid panic in case it sent a bunch of mangled text) (which it didn't 😸 ).

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -65,6 +65,13 @@ definitions:
         text: "((.:failure_snippet)) \n
               - <https://cd.gds-reliability.engineering/teams/$BUILD_TEAM_NAME/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME|Concourse build #$BUILD_NAME>"
 
+  put_db_migration_slack_notification: &put_db_migration_slack_notification
+    put: slack-notification
+    params:
+      channel: '#govuk-pay-announce'
+      text: ":postgres: FARGATE starting $BUILD_JOB_NAME on production-2\n
+            - <https://cd.gds-reliability.engineering/teams/$BUILD_TEAM_NAME/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME|Concourse build #$BUILD_NAME>"
+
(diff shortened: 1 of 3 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -65,6 +65,13 @@ definitions:
         text: "((.:failure_snippet)) \n
               - <https://cd.gds-reliability.engineering/teams/$BUILD_TEAM_NAME/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME|Concourse build #$BUILD_NAME>"
 
+  put_db_migration_slack_notification: &put_db_migration_slack_notification
+    put: slack-notification
+    params:
+      channel: '#govuk-pay-announce'
+      text: ":postgres: FARGATE starting $BUILD_JOB_NAME on production-2\n
+            - <https://cd.gds-reliability.engineering/teams/$BUILD_TEAM_NAME/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME|Concourse build #$BUILD_NAME>"
```
oswaldquek, line 73: I take it `BUILD_PIPELINE_NAME`, `BUILD_JOB_NAME`, and `BUILD_NAME` are standard env vars for that running job?

oswaldquek (approved):
LGTM. I can see this was tested at https://gds.slack.com/archives/CEBPUR2JE/p1622634492013800.

ci/pipelines/deploy-to-production.yml
katstevens: Correct ✔️

PR #500, opened 2021-06-28
PP-8162: Create failure notification snippet

oswaldquek:
The team are in agreement that just like for the
nginx/nginx-forward-proxy/telegraf deploys, we only need failure notifications
for carbon-relay deploys.

Contrary to the PR name (with "check-release-version") I've decided to go for a small commit here and do that functionality in another commit.

Example of a successful build: https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-carbon-relay/builds/21.

Example notification: https://gds.slack.com/archives/CNZFSNM0S/p1624875712015800

PR #503, opened 2021-06-28
PP-8162: Deploy carbon-relay to staging and push to prod ecr

oswaldquek:
[Successful deploy of carbon-relay to staging](https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-staging/jobs/deploy-carbon-relay/builds/6) followed by a [push of the 2-carbon-relay-release to prod ecr](https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-staging/jobs/push-carbon-relay-to-production-ecr/builds/1).

This commit also introduces an STUNNEL_IMAGE_TAG environment variable. This [build](https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-carbon-relay/builds/27) verifies the deploy-to-test pipeline still works.

### DIFF
--- a/ci/pipelines/deploy-to-staging.yml
+++ b/ci/pipelines/deploy-to-staging.yml
@@ -168,6 +168,19 @@ resources:
       branch: master
       username: alphagov-pay-ci
       password: ((github-access-token))
+  - name: carbon-relay-ecr-registry-staging
+    type: registry-image-resource-1-1-0
+    icon: docker
+    source:
+      repository: govukpay/carbon-relay
+      variant: carbon-relay-release
+      <<: *aws_staging_config
+  - name: carbon-relay-ecr-registry-prod
+    type: registry-image-resource-1-1-0
+    icon: docker
+    source:
+      repository: govukpay/carbon-relay
+      <<: *aws_production_config
   - name: toolbox-ecr-registry-staging
     type: registry-image-resource-1-1-0
     icon: docker
@@ -436,6 +449,10 @@ groups:
       - deploy-toolbox-to-staging
       - smoke-test-toolbox-on-staging
       - push-toolbox-to-production-ecr
+  - name: carbon-relay
+    jobs:
+      - deploy-carbon-relay-to-staging
+      - push-carbon-relay-to-production-ecr
   - name: telegraf
     jobs:
       - deploy-toolbox-to-staging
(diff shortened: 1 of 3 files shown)


ci/pipelines/deploy-to-staging.yml
```diff
@@ -460,7 +477,82 @@ jobs:
         trigger: true
       - set_pipeline: deploy-to-staging
         file: deploy-to-staging-pipeline-definition/ci/pipelines/deploy-to-staging.yml
-        
+
+  - name: deploy-carbon-relay
```
katstevens, line 481: Can we name this `deploy-carbon-relay-to-staging` to be consistent with the other apps?

ci/pipelines/deploy-to-test.yml
```diff
@@ -684,6 +684,7 @@ jobs:
         params:
           <<: *aws_assumed_role_creds
           CARBON_RELAY_IMAGE_TAG: ((.:carbon_relay_image_tag))
+          STUNNEL_IMAGE_TAG: 1-stunnel-release
```
katstevens, line 687: Are we just hardcoding this for the moment then?

ci/pipelines/deploy-to-staging.yml
```diff
@@ -460,7 +477,82 @@ jobs:
...
+          run:
+            path: sh
+            args:
+              - -c
+              - |
+                cat <<EOT >> snippet/failure
+                :red_circle: FARGATE Deployment of <https://github.com/alphagov/pay-dockerfiles/releases/tag/${CARBON_RELAY_IMAGE_TAG}|carbon-relay v${CARBON_RELAY_IMAGE_TAG}> on ${ENV} failed. Version details:
+                EOT
+      - load_var: failure_snippet
+        file: snippet/failure
```
katstevens, line 510: For staging app releases we usually have a success notification as well. I'm happy for this to be added in a subsequent PR though.

ci/pipelines/deploy-to-test.yml
oswaldquek: Yes

katstevens (commented):
Looking good - one small name change and I think we can merge.

ci/pipelines/deploy-to-staging.yml
oswaldquek: I thought we agreed that we would be happy for just a failure notification for the carbon-relay deploy? That's the way it is for the deploy-to-test. From https://github.com/alphagov/pay-ci/pull/500:
```
The team are in agreement that just like for the
nginx/nginx-forward-proxy/telegraf deploys, we only need failure notifications
for carbon-relay deploys.
```
It is an app and not a sidecar like nginx is though, so I can see the rationale for the start and success notifications as well. Happy to change it if you like?

ci/pipelines/deploy-to-staging.yml
oswaldquek: This has been applied: https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-staging/jobs/deploy-carbon-relay-to-staging/builds/1

ci/pipelines/deploy-to-staging.yml
oswaldquek: If we want a start and success notification it'd be easier to do in another PR.

katstevens: Hmm good point! I think the reasoning behind the auto-slack notifications is so that devs don't have to post about the release in the channel themselves. The telegraf/nginx stuff only ever goes out with an app, so we'd 'know' about it via that app's notification. Whereas the carbon relay app is its own task (I don't think we'd need separate notifications for stunnel, for example, as that's always going out with a carbon relay 'app' release). Maybe we should add start/success to the test pipeline after all, and send them to the `govuk-pay-activity` black hole?

katstevens: (To reiterate, I don't think we need to do it in this PR)

katstevens (approved):

PR #505, opened 2021-06-29
PP-8162: Deploy carbon-relay to production

oswaldquek:
[Succesful deploy to prod](https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-production/jobs/deploy-carbon-relay-to-prod/builds/2).

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -34,7 +34,7 @@ definitions:
     aws_role_arn: arn:aws:iam::((pay_aws_prod_account_id)):role/concourse
     aws_ecr_registry_id: "((pay_aws_prod_account_id))"
     aws_region: eu-west-1
-    
+
   put_start_slack_notification: &put_start_slack_notification
     put: slack-notification
     params:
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -346,6 +356,60 @@ groups:
...
+          AWS_REGION: "eu-west-1"
+          CLUSTER_NAME: "production-2-fargate"
+          APP_NAME: carbon-relay
+          CARBON_RELAY_IMAGE_TAG: ((.:carbon_relay_image_tag))
+      - task: deploy-to-prod
+        file: pay-ci/ci/tasks/deploy-carbon-relay.yml
+        params:
+          <<: *aws_assumed_role_creds
+          CARBON_RELAY_IMAGE_TAG: ((.:carbon_relay_image_tag))
+          STUNNEL_IMAGE_TAG: test
```
katstevens, line 400: So this would be `0-release` at the moment for prod

oswaldquek: I noticed I hadn't committed it at the same time you commented! Pushed now.

ci/pipelines/deploy-to-production.yml
```diff
@@ -346,6 +356,60 @@ groups:
       - update-deploy-to-prod-pipeline
 
 jobs:
+  - name: deploy-carbon-relay-to-prod
```
katstevens, line 359: I think positioning it first in the `jobs` list affects the ordering of the tabs in the pipeline - we should prob keep it consistent with staging and test and stick it between toolbox and the sidecars.

oswaldquek: Fixed

katstevens (approved):

PR #515, opened 2021-07-28
PP-8166 Add Docker to ECR for tcp-proxy

danworth:
This will trigger when a new release is created for the tpc-proxy, pull
the image form Docker hub and push into test ECR, just like we for the
other images being built by Jenkins.

## WHAT
Applied to `deploy-to-test` pipeline and it works: https://cd.gds-reliability.engineering/teams/pay-dev/pipelines/deploy-to-test/jobs/push-tcp-proxy-to-test-ecr/builds/1


dj-maisy (approved):
Can't see any issues with this. 👍

PR #530, opened 2021-08-25
PP-8547 Fix egress deploy timeout

katstevens:
Addresses the timeout issue seen in https://cd.gds-reliability.engineering/teams/pay-deploy/pipelines/deploy-to-production/jobs/deploy-egress-to-prod/builds/3.

Egress sits behind a Network Load Balancer, which has a longer wait time for deregistering the old tasks (which means less time for actual deploying). I've increased it to 15 minutes.

Also:
- fixed the timeout log message which was showing seconds instead of milliseconds
- committed some linting suggestions from running `npm run lint` (StandardJS).


oswaldquek (approved):
LGTM. Feel free to ignore, but it might be better to convert the `Deployment did not complete after...` message to minutes and seconds.

PR #537, opened 2021-09-30
PP-8711: Upgrade terraform version to 0.14.11

katstevens:
We've upgraded the deploy terraform in pay-infra, which is now running on 0.14.11.

This PR upgrades the Concourse pipelines to do the same (fetching the registry image from hashicorp/terraform).


jfharden (approved):
Oh cool, didn't realise there was this.

PR #539, opened 2021-09-30
PP-8711: Upgrade to terraform 1.0.8

jfharden:
Terraform is being upgraded to 1.0.8 for all microservices in PR https://github.com/alphagov/pay-infra/pull/3313

This PR needs to be merged prior to https://github.com/alphagov/pay-infra/pull/3313 is applied to enable deploys to continue.


jfharden:
I've just added the deploy-smoke-tests pipeline in this update too. The version lock on that is "~> 1.0" so it doesn't require any terraform change to accompany it.

katstevens (approved):
Ready when we're ready 👍

PR #553, opened 2021-11-23
BAU: Fix shellcheck errors

jfharden:
There are a lot of shellcheck erros in the ci for this repo, most of them are trivial to fix. Hopefully this PR should produce a green build and also still work everywhere.

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -1351,7 +1351,7 @@ jobs:
         outputs:
           - name: pacts
         run:
-          path: sh
+          path: bash
           dir: src
           args:
             - -ec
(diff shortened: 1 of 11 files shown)


katstevens:
Just a warning that the last time I tried fixing this (in Feb 2021), it broke the pact tests. [I used `bash` instead of `ash`](https://github.com/alphagov/pay-ci/pull/368) though so hopefully it'll work this time?

jfharden:
@katstevens was that in the zip-build task? I don't suppose you remember, I've changed zip-build to ash, and changed a different (also pact related) task to bash

jfharden:
As you noticed in your PR of course even with all the shellcheck errors gone the concourse pipelines still fail their validation :(

jfharden:
There are only the concourse errors left.

I propose: Until we schedule real time to fix them we disable them and replace them with the simpler `fly validate` with fly cli 7.6.0. This will not tell us about any of the kinds of things pipecleaner does, but we aren't even looking at them because the tests are always dead. I think it's better to have a less good but actually working test than a completely ignored test suite. This also means with the shellcheck fixes we would be actually alerted to failures

katstevens:
> @katstevens was that in the zip-build task? I don't suppose you remember, I've changed zip-build to ash, and changed a different (also pact related) task to bash

Annoyingly it was long enough ago that the build history has fallen off Big Concourse. zip-build sounds likely though.

katstevens (approved):
Happy to try this, however based on past experience, I'd be ready with a revert PR just in case there's something weird buried in a task that fails...

jfharden:
I'm going to merge this as-is while most people aren't in work yet, we can revert if need be. I'm going to manually run a bunch of the thinks I've changed with this change to make sure they are ok.

jfharden:
After merge and pipeline update I've manually run at least one job that uses every one of the update tasks and jobs, they all pass ok. The only thing is it means it hasn't actively published any new pact tags since there's nothing new to run though. I'm going to keep an eye out for anything running which will. I'm pretty confident this will be ok though.

PR #554, opened 2021-11-24
BAU: Limit pr-ci tests so they pass and are not ignored

jfharden:
Limit pr-ci tests.

The tests for the pr-ci pipeline have been failing since before I started, this means they are always ignored and any new potential issues are not flagged. In #553 I fixed all the shellcheck errors, this still leaves us with pipeline validation fails. The tool used to do this test is called pipecleaner and is a gds written tool. It doesn't have the ability to say treat warnings as ok, any warning causes a failure.

To make the tests there are more useful, and to make new shellcheck failures actually get flagged and noticed I've changed the validation to just use fly validate for the pipeline yaml files, this does print out some deprecations but they are not treated as an error, and use pipecleaner to validate the individual task yamls which all pass already.

In theory this is worse since pipecleaner checks more, but since they have been failing for a long time, we haven't done anything about them, and don't really have time (or inclination if we aren't going to stick with concourse long term) to fix them it's better to limit the tests to a successful set which will flag up some new problems if they occur and always show new shellcheck problems. Given most of the team aren't bash expects i think the shellcheck errors are especially useful.

Note: I set the pipeline manually (verifying the changes were only those that were made here, and you can see the result is a test pass: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/ci-pr-test/builds/12

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -1952,9 +1952,15 @@ jobs:
               apk add git shellcheck
               go get github.com/alphagov/paas-cf/tools/pipecleaner
 
-              wget -c https://github.com/concourse/concourse/releases/download/v7.6.0/fly-7.6.0-linux-amd64.tgz -O - | \
-                tar -O -zxf - > /usr/local/bin/fly
+              cd /tmp/
+
+              echo "c7d331052a6bf552b017adf5288b8e162346157c  fly-7.6.0-linux-amd64.tgz" > fly-7.6.0-linux-amd64.tgz.sha1
+              wget -c https://github.com/concourse/concourse/releases/download/v7.6.0/fly-7.6.0-linux-amd64.tgz -O fly-7.6.0-linux-amd64.tgz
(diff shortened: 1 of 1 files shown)


ci/pipelines/pr.yml
```diff
@@ -1951,7 +1951,18 @@ jobs:
             - |
               apk add git shellcheck
               go get github.com/alphagov/paas-cf/tools/pipecleaner
-              pipecleaner --rubocop=false ci/pipelines/*.yml ci/tasks/*.yml
+
+              wget -c https://github.com/concourse/concourse/releases/download/v7.6.0/fly-7.6.0-linux-amd64.tgz -O - | \
```
jfharden, line 1955: I should validate the sha of this too

jfharden: Done

katstevens (approved):
This looks great. It's definitely an improvement over our previous suggested option of 'turn off this check entirely'. The Concourse job output is easier to read as well as it's broken down by pipeline.

PR #555, opened 2021-11-26
Experiment: Add get-deployed-image-tags script

jfharden:


ci/scripts-aws-sdk-3/get-deployed-image-tags.js
```diff
@@ -0,0 +1,95 @@
...
+  const taskDefMap = new Map()
+
+  const promises = []
+
+  // We can only request 10 service descriptions in a single
+  // call, so chunk it up
+  while (services.length > 0) {
+    promises.push(
+      ecsClient.describeServices({
+        cluster: 'test-12-fargate',
```
danworth, line 40: Just FYI in case you've forgotten the cluster is hard-coded here.

ci/scripts-aws-sdk-3/get-deployed-image-tags.js
```diff
@@ -0,0 +1,95 @@
...
+          repository: containerDef.image.split(delimiter)[0],
+          tag: containerDef.image.split(delimiter)[1]
+        }
+      }
+    )
+  }
+}
+
+async function describeServices (services) {
+  const taskDefMap = new Map()
```
danworth, line 31: This is more personal choice and left here as a musing, I prefer declaring variables close to where they're used rather than at the start of the scope as I feel it helps readability and refactoring later on, as such I would have put `const taskDefMap = new Map()` on line 47.

danworth (commented):
I think that's some nice async/await node.js and makes good use of Node's async features 🏅  I found it simple to understand and follow. As mentioned in the Slack thread `Jest` is a nice testing framework and there's a useful AWS SDK v3 mocking library available.

jfharden:
This is not required

PR #556, opened 2021-11-30
Delete node-build.yml

0atman:
~~This file was for travis - we don't use travis anymore.~~

This file is unused


katstevens:
Hi @0atman - I think this file is still used in the Concourse PR pipeline?

0atman:
@katstevens but it's travis, isn't it?

0atman:
I'm no longer convinced this file is travis :laughing:

0atman:
OK, it's certainly concourse, we were bamboozled by it being added at the same time as a travis file

0atman:
But do we use it? :confused:

jfharden:
I think this can be deleted @katstevens the file used in the pr pipeline is node-build-pr.yml https://github.com/alphagov/pay-ci/blob/master/ci/pipelines/pr.yml#L199 and I can't find any reference to node-build.yml in any pipeline in this repo, pay-team-manual, pay-tech-docs or pay-js-commons

0atman:
:zombie:

katstevens:
@jfharden Ah you're right - I was getting confused with the `node-build-pr.yml`. Now we've got the pipeline verification check we should have more confidence in deleting this, right?

iqbalgds (approved):

jfharden:
Sadly likely not, if the pipeline refers to a file then it doesn't check to see if that file exists since it's not cloning the other repos it requires. However I'm confident this isn't used and the worst side effect is a step failing in a new build, at which point we can trivially revert this PR since it's so well focussed and just allow the pipeline to be set again automatically.

jfharden (approved):

PR #557, opened 2021-11-30
Update alpine version and pin node and octorest to major versions

0atman:
Driveby update of alpine while we pinned the aws-sdk and octoreset

### DIFF
--- a/ci/docker/node-runner/Dockerfile
+++ b/ci/docker/node-runner/Dockerfile
@@ -1,4 +1,5 @@
-FROM node:12.22.4-alpine3.14@sha256:56c0e7c15a9cfb4359d463912ad0b54da4e73e917f00fd4fc8d73c67a93252a2
+FROM node:12.22.4-alpine3.12@sha256:a199f64c814f12fb00fa81a6e1c7f964e260a127f2b64ac526a358f3152c0f66
 
 RUN npm install aws-sdk@^2.x.x
 RUN npm install @octokit/rest@^18.x.x
+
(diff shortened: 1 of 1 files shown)


0atman:
@katstevens very reasonable! I'm on a call with @jfharden, who suggested this, but we'll cool our jets for now :rocket:

ci/docker/node-runner/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM node:12.22.4-alpine3.12@sha256:a199f64c814f12fb00fa81a6e1c7f964e260a127f2b64ac526a358f3152c0f66
```
0atman, line 1: @katstevens reverted the alpine changes

ci/docker/node-runner/Dockerfile
```diff
@@ -1,5 +1,5 @@
...
 
-RUN npm install aws-sdk
-RUN npm install @octokit/rest
+RUN npm install aws-sdk@^2.x.x
+RUN npm install @octokit/rest@^18.x.x
```
0atman, line 4: @jfharden checked the live versions of both these, and we've pinned to them.

jfharden:
We know alpine 3.12 has critical vulnerabilities, but as it happens so did the image built of this container, sadly though I think you're right @katstevens and we need to make a ticket for this.

We can at least lock the aws-sdk and octocat/rest versions to the current major versions they have to ensure a rebuild of this image doesn't break our current usage of this container.

jfharden:
Speaking with @katstevens I'm quite nervous about building a new version of this container without doing some serious detective work to find where it's used and test it well first (there's no testing whatsoever for this container at the moment).

However this specific change helps future updaters not break anything, and we have no auto build pipeline so I propose to merge this PR as-is, but not to build a new version of the container yet.

We have a story in our backlog to update the alpine version of this container which will give us chance to do a more extensive test (and hopefully write some automated tests).

jfharden (approved):
Approving this on the basis of just merging the PR and not building a new version of the container.

We have Jira story PP-8980 to update alpine at which point we can invest some time into testing and likely automated testing of the container.

0atman:
Merged, this should have no effect on the two npm install commands as they pin them to the current major versions :ok_hand:

0atman:
Thanks all for your help and advice!

PR #559, opened 2021-12-10
PP-8979: Use concourse/registry-image-resource v1.4.1

oswaldquek:
Using the latest stable version of the resource may help address some issues
we've seen with jobs not triggering correctly.


dj-maisy (approved):
Giving this a 🟢 , but wait until the deployment freeze is removed.

oswaldquek:
> Giving this a 🟢 , but wait until the deployment freeze is removed.

It's already been applied via `set-pipeline`.

PR #561, opened 2021-12-14
BAU: Fix products slack notifications

katstevens:
Spotted two typos - one for `products` smoke test (appearing as `ledger` in Slack) and one for `products-ui` (appearing as `products` in Slack).


SandorArpa (approved):
LOKTM

PR #565, opened 2022-01-06
PP-8980: Upgrade to alpine 3.14

oswaldquek:


ci/docker/node-runner/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12.22.4-alpine3.12@sha256:a199f64c814f12fb00fa81a6e1c7f964e260a127f2b64ac526a358f3152c0f66
+FROM node:12.22.4-alpine3.14@sha256:a199f64c814f12fb00fa81a6e1c7f964e260a127f2b64ac526a358f3152c0f66
```
jfharden, line 1: You need to update the sha, the actual tag is ignored when providing an image sha

jfharden: Just to add a bit of context, we keep the tag in the Dockerfile even though it's ignored since it makes it a _lot_ easier to work out what tag we were aiming for :)

oswaldquek: Ok, I've used the sha from https://hub.docker.com/layers/alpine/library/alpine/3.14/images/sha256-5e604d3358ab7b6b734402ce2e19ddd822a354dc14843f34d36c603521dbb4f9?context=explore.

dj-maisy: That's a SHA for amd64, so that should be good. 👍

jfharden: I've just realised, that's the wrong image, this container is built on the node image rather than the alpine one (but an alpine tagged node image).

image name is node
image tag is 12.22.4-alpine3.14

So I think you want https://hub.docker.com/layers/node/library/node/12.22.4-alpine3.14/images/sha256-56c0e7c15a9cfb4359d463912ad0b54da4e73e917f00fd4fc8d73c67a93252a2?context=explore

I nearly missed that, the hashes are _really_ close to each other O_o

oswaldquek: Thanks, updated.

jfharden (approved):
👍

PR #567, opened 2022-01-11
PP-9096: Clarify the job formerly known as 'pact-provider-verification'

oswaldquek:
Connector and ledger PR builds contain a 'pact-provider-verification' job, in
addition to a 'pact-provider-test' job which other app PR builds have. This is
because connetor and ledger are both consumers and providers in the pact world.
The 'pact-provider-verification' job actually runs pact provider tests where
they are the consumer. This PR gives clarity to this purpose. We also set the
context as 'app-as-consumer pact verification' which is used by the GitHub PR
as an additional check.

Naturally, we could clarify the 'pact-provider-test' job and rename it to
'app-as-provider pact verification' but that will come in another PR.

Test builds which demonstrate this:
https://github.com/alphagov/pay-connector/pull/3397
https://github.com/alphagov/pay-ledger/pull/1579

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -338,7 +338,7 @@ groups:
       - card-connector-unit-test
       - card-connector-integration-test
       - card-connector-pact-provider-test
-      - card-connector-pact-provider-verification
+      - card-connector-as-consumer-pact-test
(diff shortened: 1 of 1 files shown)


stephencdaly (approved):
Looks good to me

PR #568, opened 2022-01-11
PP-9096: Rename <app>-pact-provider-test job

oswaldquek:
Given we have renamed the `<app>-pact-provider-verification` job to
`<app>-as-consumer-pact-test` (see
https://github.com/alphagov/pay-ci/pull/567), it makes sense to rename
`<app>-pact-provider-test` to `<app>-as-provider-pact-test` for clarity.


katstevens (approved):
😻 Another step towards me actually getting my head around the pact tests - thank you!

PR #569, opened 2022-01-11
bau: Address deprecation warnings

oswaldquek:
Amending as per the rules at https://concourse-ci.org/config-basics.html#schema.identifier.

There are way more identifiers to rectify but thought I'd raise this as a preliminary pass for people to have a look at.


jfharden (approved):
Looks like a good start 👍

PR #570, opened 2022-01-12
PP-9107 Add webhooks deploy to test job

mrlumbu:
First step to creating pipeline for webhooks.
This is currently just the deploy job.
It will be triggered by manual pushes to
ECR for now. I have not included other
jobs (pushes to ecr, smoke tests) as they
will not currently work.

For deploy job itself I have removed
pact tagging and release version checking
to allow any manual tag to be deployed.

I have added a different variant to the
webhook ecr resource. This means it will
look for images of type 123-test_release.
This means we won't pollute the world
with dodgy release numbers when doing
manual testing, and can switch to
standard format when these release
numbers are coming from github.

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -265,6 +265,13 @@ resources:
       repository: govukpay/ledger
       variant: release
       <<: *aws_test_config
+  - name: webhooks-ecr-registry-test
+    type: registry-image
+    icon: docker
+    source:
+      repository: govukpay/webhooks
+      variant: test_release
+      <<: *aws_test_config
   - name: products-ecr-registry-test
     type: registry-image
     icon: docker
@@ -553,6 +560,9 @@ groups:
       - deploy-toolbox
       - smoke-test-toolbox
       - push-toolbox-to-staging-ecr
+  - name: webhooks
+    jobs:
+      - deploy-webhooks
   - name: carbon-relay
     jobs:
       - deploy-carbon-relay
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -553,6 +560,9 @@ groups:
       - deploy-toolbox
       - smoke-test-toolbox
       - push-toolbox-to-staging-ecr
+  - name: webhooks
+    jobs:
+      - deploy-webhooks
```
katstevens, line 565: In theory there's nothing to stop you adding a `smoke-test-webhooks` here (and copy-paste the definition further down obv). For example we run the Canaries against test-12 after Toolbox release, even though they don't have coverage for Toolbox. Then if something weird *does* happen, at least you'll get a notification...

ci/pipelines/deploy-to-test.yml
```diff
@@ -2770,6 +2780,61 @@ jobs:
...
+      - get: webhooks-ecr-registry-test
+        trigger: true
+      - get: nginx-proxy-ecr-registry-test
+        trigger: true
+      - get: telegraf-ecr-registry-test
+        trigger: true
+      - get: pay-infra
+      - get: pay-ci
+      - load_var: application_image_tag
+        file: webhooks-ecr-registry-test/tag
```
katstevens, line 2797: I think this file won't exist without the equivalent `parse-release-tag` job - you might have to hardcode or do some wrangling

ci/pipelines/deploy-to-test.yml
```diff
@@ -2770,6 +2780,61 @@ jobs:
...
+          <<: *snippet_params_all_versions
+      - load_var: success_snippet
+        file: snippet/success
+      - load_var: failure_snippet
+        file: snippet/failure
+      - task: assume-role
+        file: pay-ci/ci/tasks/assume-role.yml
+        params:
+          AWS_ROLE_ARN: arn:aws:iam::((pay_aws_test_account_id)):role/concourse
+          AWS_ROLE_SESSION_NAME: terraform-staging-assume-role
```
katstevens, line 2816: I think this should be `terraform-test-assume-role`?

ci/pipelines/deploy-to-test.yml
mrlumbu: pretty sure this doesn't rely on any other job (or step), it's just relying on the file created by the resource, as documented here https://github.com/concourse/registry-image-resource#files-created-by-the-resource

katstevens: Ah you're right. What sort of tag do we get there (i.e. at some stage will we have to do some formatting to get a nice `123-release` for use further down the pipeline)?

mrlumbu: yep, the resource dictates which tags it will trigger on here https://github.com/alphagov/pay-ci/blob/867aa56d7acaf262af50b1ff56059f2e0c3ca6d9/ci/pipelines/deploy-to-test.yml#L273

At the moment I have it listening for test_release-XXX so we can deploy stuff like that without polluting the `release-XXX` space, which might get confusing given downgrade protection etc. So at the moment this will deploy any tag that is pushed to ecr webhooks of form `test_release-XXX`

ci/pipelines/deploy-to-test.yml
mrlumbu: think we'll leave these off just for now - as you say, easy to add in when needed

katstevens (approved):
OK - happy to give this a whirl

PR #572, opened 2022-01-13
BAU: Make deployment notifications less shouty

katstevens:
Also changes the icon to be more relevant to what's happening, eg Fargate or Postgres.

Following feedback from our users that shouting 'FARGATE' in big letters on each notification is no longer necessary.


alexbishop1 (approved):
image

katstevens:
Missed a few jobs - pushed a commit to fix

alexbishop1 (approved):

PR #573, opened 2022-01-14
BAU: Allow rollback of connector in test and staging

jfharden:
In an emergency we allowed staging and test to rollback to an older version (2578-release). This has been applied and should be reverted after the weekend


katstevens:
@jfharden if this has been reverted, can we close this PR?

PR #577, opened 2022-01-24
PP-9107 Build webhooks job

mrlumbu:
Triggers on an appropriately formatted git tag, and then builds jar and
from that the docker image.
This is a pure build step - all tests should have been run prior to this
(except e2e, which is out of scope for now).
Eventually build jar and build image tasks can be extracted, but just
want to leave them in line for now in case we want to fiddle with them

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -197,6 +197,13 @@ resources:
     source:
       uri: https://github.com/alphagov/pay-selfservice
       branch: master
+      tag_regex: "alpha_release-(.*)"  
+  - name: webhooks-git-release
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-webhooks
+      branch: main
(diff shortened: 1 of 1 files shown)


gidsg (approved):
👍  Code LGTM (although I don't know how to actually test this on Concourse)

ci/pipelines/deploy-to-test.yml
```diff
@@ -2787,6 +2795,91 @@ jobs:
...
+              repository: govukpay/concourse-runner
+              tag: latest
+          inputs:
+            - name: jar
+          outputs:
+            - name: local_image
+          params:
+            app_name: webhooks
+          run:
+            path: bash
```
katstevens, line 2849: Guessing this section is based on the `build-docker-image.yml` task, with a couple of tweaks? Could we reuse that task going forward?

katstevens (approved):
Looks good - makes sense to keep the release tagging/parsing consistent with the other apps for now.

PR #578, opened 2022-02-01
Pp 9086 check for bastion update

jfharden:
Create pipelines for test, staging, and production which check to see if the bastion plan has any changes.

This will alert us when there is a new ami available since drift will be detected. As a positive side effect we will also see if anyone changes anything without our knowledge.

See https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/infra-drift-detector for the lower-envs file having been applied. Only test-12 will work right now.

There is an accompanying PR to create the role used for this in pay-infra https://github.com/alphagov/pay-infra/pull/3545

This has only been applied in test-12 so the assume-role step will fail right now in all other jobs other than test-12.

### DIFF
--- a/ci/pipelines/infra-drift-detector-higher-envs.yml
+++ b/ci/pipelines/infra-drift-detector-higher-envs.yml
@@ -57,7 +57,7 @@ jobs:
     - task: assume-role
       file: pay-ci/ci/tasks/assume-role.yml
       params:
-        AWS_ROLE_ARN: arn:aws:iam::((pay_aws_staging_account_id)):role/pay-concourse-bastion-staging-2
+        AWS_ROLE_ARN: arn:aws:iam::((pay_aws_staging_account_id)):role/pay-concourse-bastion-read-only-staging-2
         AWS_ROLE_SESSION_NAME: pay-concourse-bastion-drift-detection
     - load_var: role
       file: assume-role/assume-role.json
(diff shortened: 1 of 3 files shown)


ci/pipelines/infra-drift-detector-lower-envs.yml
```diff
@@ -0,0 +1,145 @@
...
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: master
+  - name: slack-notification
+    type: slack-notification
+    source:
+      url: https://hooks.slack.com/services/((slack-notification-secret))
+
+definitiions:
```
katstevens, line 42:
```suggestion
definitions:
```

ci/pipelines/infra-drift-detector-higher-envs.yml
```diff
@@ -0,0 +1,145 @@
...
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: master
+  - name: slack-notification
+    type: slack-notification
+    source:
+      url: https://hooks.slack.com/services/((slack-notification-secret))
+
+definitiions:
```
katstevens, line 42:
```suggestion
definitions:
```

katstevens (commented):
Looks good - a couple of tiny typos. However see my comment in the other PR about the assumed role name (not a blocker though)

ci/pipelines/infra-drift-detector-lower-envs.yml
jfharden: I am surprised this works O_o

Not only works, but works! and doesn't error with a concourse validation or set-pipeline O_o whuuuuu

jfharden: Interestingly this is totally ignored whether I change it or not, so maybe concourse ignores all the top level keys it doesn't know

jfharden: Sorted both of them, thanks! 👍

katstevens (approved):
* hums 'Up on the Roof' by The Drifters *

PR #580, opened 2022-02-02
PP-9086: Fix production account variable name

jfharden:
The prod account id variable name is prod, not production. This has been applied and works: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/infra-drift-detector/jobs/production-2-bastion/builds/2


katstevens (approved):
Good spot!

PR #583, opened 2022-02-16
PP-9231: Specify docker registry uri for e2e tests

jfharden:
In preparation for running end to end tests in CodeBuild it needs to be possible to specify a docker registry which is not dockerhub.

This change allows specifying the env var DOCKER_REGISTRY_URI to use a registry other than dockerhub.

Note the syntax `${DOCKER_REGISTRY_URI:-}` means to default to a blank string if it's unset meaning where it's not set in jenkins/concourse it will continue to use the dockerhub registry.

# How:

To test them like jenkins would run them with no env var run them locally (warning, this will take a WHILE (maybe 15 mins+) and will pull several gigs of containers from dockerhub:
```
$ cd ci/tasks/endtoend/card
$ docker-compose up --force-recreate -d
$ endtoend=$(docker ps -aqf "name=endtoend")
$ docker exec ${endtoend} /app/bin/e2e-card
```

Also to see how they run when the env var is set this build was run using these compose files and setting an ECR registry:
https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/codebuild-e2e/jobs/frontend/builds/167

### DIFF
--- a/ci/tasks/endtoend/card/docker-compose.yml
+++ b/ci/tasks/endtoend/card/docker-compose.yml
@@ -6,7 +6,7 @@ networks:
 
 services:
   publicapi:
-    image: govukpay/publicapi:${tag_publicapi:-latest-master}
+    image: ${DOCKER_REGISTRY_URI:-}govukpay/publicapi:${tag_publicapi:-latest-master}
     env_file:
       - ../docker-config/java_app.env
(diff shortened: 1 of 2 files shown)


katstevens (approved):
Guessing we'll start using this in part 2 😸

PR #584, opened 2022-02-16
PP-9231: Add run-codebuild ci script

jfharden:
Note: Over 5800 lines of this is the package.lock.json!

Add a script that can be used in a concourse pipeline to execute a codebuild and return logs almost live as the build progesses.

I've also added a workflow to run both a lint check (using standard.js) and unit tests on any changes in the `ci/scripts/run-codebuild` directory.

I've done this in a sub-directory of ci/scripts to isolate it's package from the others allowing me to use aws-sdk v3 among other things and not have dependency pollution between different scripts. I modified the gitignore to ignore node_modules in ci/scripts and all subdirectories.

For information about running it see the README included in this PR.

Also see it in action here: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/codebuild-e2e/jobs/frontend/builds/167

The run-codebuild-product and run-codebuild-card tasks execute this script.

### DIFF
--- a/.github/workflows/test-run-codebuild.yml
+++ b/.github/workflows/test-run-codebuild.yml
@@ -17,13 +17,20 @@ jobs:
     env:
       working-directory: ci/scripts/run-codebuild
 
+    strategy:
+      matrix:
+        node-version:
+          - "12.22.7"
+          - "14.19.0"
+          - "16.14.0"
+
     steps:
       - name: Checkout
         uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
-      - name: Setup
+      - name: Setup Node.js
         uses: actions/setup-node@1f8c6b94b26d0feae1e387ca63ccbdc44d27b561
         with:
-          node-version: 12.22.7
+          node-version: ${{ matrix.node-version }}
       - name: Install dependencies
         run: npm ci
         working-directory: 'ci/scripts/run-codebuild'
--- a/.github/workflows/test-run-codebuild.yml
+++ b/.github/workflows/test-run-codebuild.yml
@@ -1,0 +1,35 @@
+name: Test run-codebuild ci script
+
+on:
+  pull_request:
+    paths:
+      - 'ci/scripts/run-codebuild/**'
+  push:
+    branches:
+      - main
+    paths:
+      - 'ci/scripts/run-codebuild/**'
+
+jobs:
+  test-script-run-codebuild:
+    runs-on: ubuntu-20.04
+    name: Test script run-codebuild
+    env:
+      working-directory: ci/scripts/run-codebuild
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+      - name: Setup
+        uses: actions/setup-node@1f8c6b94b26d0feae1e387ca63ccbdc44d27b561
+        with:
+          node-version: 12.22.7
+      - name: Install dependencies
+        run: npm ci
+        working-directory: 'ci/scripts/run-codebuild'
+      - name: Lint check
+        run: npm run lint
+        working-directory: 'ci/scripts/run-codebuild'
+      - name: Unit tests
+        run: npm test
+        working-directory: 'ci/scripts/run-codebuild'
--- a/.github/workflows/test-run-codebuild.yml
+++ b/.github/workflows/test-run-codebuild.yml
@@ -6,7 +6,7 @@ on:
       - 'ci/scripts/run-codebuild/**'
   push:
     branches:
-      - main
+      - master
     paths:
       - 'ci/scripts/run-codebuild/**'
 
--- a/.github/workflows/test-run-codebuild.yml
+++ b/.github/workflows/test-run-codebuild.yml
@@ -21,8 +21,8 @@ jobs:
       matrix:
         node-version:
           - "12.22.7"
-          - "14.19.0"
-          - "16.14.0"
+          - "14"
+          - "16"
 
     steps:
       - name: Checkout
--- a/ci/scripts/run-codebuild/README.md
+++ b/ci/scripts/run-codebuild/README.md
@@ -1,0 +1,39 @@
+# Run Codebuild
+
+This script executes a CodeBuild build, allowing override of the source (and secondary source) versions, as well as the environment variables of the build. It will execute the build and then retrieve the logs from CloudWatch and print them to stdout.
+
+This script expects to be run with an environment variable `PATH_TO_CONFIG` set which specifies a path relative (or an absolute path) to the `run-codebuild.js` script. This config defines which codebuild project to run and what params to pass.
+
+## Environment Variables
+
+variable | description
+---|---
+PATH\_TO\_CONFIG | A path (absolute, or relative to the run-codebuild.js script) to the JSON config file
+
+## Config
+
+Simple example config
+```
+{
+    "projectName": "codebuild-e2e-spike-test-12",
+    "sourceVersion": "123456",
+    "secondarySourcesVersions": {
+      "pay_frontend": "654321",
+      "pay_foobar": "987654"
+      
+    },
+    "environmentVariables": {
+        "PAY_FRONTEND_RELEASE": "1734-release",
+        "END_TO_END_TEST_SUITE": "products"
+    }
+}
+```
+
+The fields in this are:
+
+field | description
+---|----
+projectName | The AWS CodeBuild project name to run
+sourceVersion | The source version for the primary source of the CodeBuild project
+secondarySourcesVersions | A map with keys for the secondary source names and values as the versions to use for those sources
+environmentVariables | A map of with keys as environment variable names to set and values as the values of those environment variables
--- a/ci/scripts/run-codebuild/README.md
+++ b/ci/scripts/run-codebuild/README.md
@@ -35,5 +35,5 @@ field | description
 ---|----
 projectName | The AWS CodeBuild project name to run
 sourceVersion | The source version for the primary source of the CodeBuild project
-secondarySourcesVersions | A map with keys for the secondary source names and values as the versions to use for those sources
-environmentVariables | A map of with keys as environment variable names to set and values as the values of those environment variables
+secondarySourcesVersions | A JSON object, the attribute names are the secondary source names, the attribute values are the versions to use for those secondary sources
+environmentVariables | A JSON object, the attribute names are the environment variable names to set, the attribute values are the values of those environment variables
--- a/ci/scripts/run-codebuild/libs/codebuild-runner.js
+++ b/ci/scripts/run-codebuild/libs/codebuild-runner.js
@@ -1,0 +1,158 @@
+// Sigh, we could use the paginator, however the paginator for GetLogEvents never completes so we have to handle it
+// ourselves and go back to the past and do all the pagination
+import { CloudWatchLogsClient, GetLogEventsCommand } from '@aws-sdk/client-cloudwatch-logs'
+import { CodeBuildClient, StartBuildCommand, BatchGetBuildsCommand } from '@aws-sdk/client-codebuild'
+
+import { promisify } from 'util'
+
+const sleep = promisify(setTimeout)
+
+export class CodeBuildError extends Error {
+  constructor (message, buildInfo) {
+    super(message)
+    this.name = 'CodeBuildError'
+    this.buildInfo = buildInfo
+  }
+}
+
+export class CodeBuildRunner {
+  constructor (projectName, sourceVersion, secondarySourceVersions, environmentVariables, region = 'eu-west-1', defaultSleepTime = 5000) {
+    this.defaultSleepTime = defaultSleepTime
+
+    this.buildCommandInput = {
+      projectName: projectName,
+      sourceVersion: sourceVersion,
+      secondarySourcesVersionOverride: secondarySourceVersions,
+      environmentVariablesOverride: environmentVariables
+    }
+
+    this.codeBuildClient = new CodeBuildClient({ region: region })
+    this.cloudWatchLogsClient = new CloudWatchLogsClient({ region: region })
+
+    this.buildId = null
+  }
+
+  async runBuildAndLog () {
+    let build = await this._startBuild()
+
+    console.log(`Sent command to start build ${build.buildNumber} for ${build.projectName} at ${build.startTime}`)
+
+    await this._waitForBuildToProduceLogs()
+
+    console.log(`Build ${this.buildId} is running, logs follow`)
+    console.log('='.repeat(80))
+
+    await sleep(this.defaultSleepTime)
+
+    build = await this._logWhileWaitingForBuildToComplete()
+
+    await sleep(this.defaultSleepTime)
+
+    console.log('='.repeat(80))
+    console.log(`Build finished with status ${build.buildStatus}`)
+
+    return await this.getBuildInfo()
+  }
+
+  async getBuildInfo () {
+    const command = new BatchGetBuildsCommand({ ids: [this.buildId] })
+
+    const commandResponse = await this.codeBuildClient.send(command)
+
+    return commandResponse.builds[0]
+  }
+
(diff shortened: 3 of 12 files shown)


.github/workflows/test-run-codebuild.yml
```diff
@@ -0,0 +1,35 @@
+name: Test run-codebuild ci script
+
+on:
+  pull_request:
+    paths:
+      - 'ci/scripts/run-codebuild/**'
+  push:
+    branches:
+      - main
```
katstevens, line 9: We're still on `master` on pay-cd 😿

.github/workflows/test-run-codebuild.yml
```diff
@@ -0,0 +1,35 @@
...
+    env:
+      working-directory: ci/scripts/run-codebuild
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+      - name: Setup
+        uses: actions/setup-node@1f8c6b94b26d0feae1e387ca63ccbdc44d27b561
+        with:
+          node-version: 12.22.7
```
katstevens, line 26: We can take advantage of the matrix testing here to do node 14 as well (given 12 is EOL very soon). Example here https://github.com/alphagov/pay-logging-lambdas/blob/main/.github/workflows/ci.yml#L7

ci/scripts/run-codebuild/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "Run codebuild script",
+  "type": "module",
+  "scripts": {
+    "test": "jest --silent=true",
+    "lint": "standard"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@aws-sdk/client-codebuild": "^3.45.0",
+    "@aws-sdk/client-cloudwatch-logs": "^3.45.0"
```
katstevens, line 13: 💯 for using specific client libraries, so pleased this is possible these days

.github/workflows/test-run-codebuild.yml
jfharden: I could add 16 as well and see if it passes :) If so win! If not I can leave it out :)

ci/scripts/run-codebuild/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "run-codebuild",
+  "version": "0.0.1",
+  "description": "Run codebuild script",
+  "type": "module",
+  "scripts": {
+    "test": "jest --silent=true",
+    "lint": "standard"
+  },
```
katstevens, line 9: Optional, but we could add an `engines` attribute here to manage expectations of what the code can run against, something like

```
 "engines": {
    "node": ">=12.22.7 <15"
  },
```

ci/scripts/run-codebuild/libs/codebuild-runner.js
```diff
@@ -0,0 +1,158 @@
...
+    while (true) {
+      const build = await this.getBuildInfo()
+
+      if (build.buildStatus !== 'IN_PROGRESS') {
+        throw new CodeBuildError(`Codebuild finished with status ${build.buildStatus} before producing any logs!`, build)
+      }
+
+      if (typeof build.logs.groupName !== 'undefined') {
+        // Newline gives us a blank line after all the progress .'s
+        console.log(`\nLogs are now availabile at ${build.logs.deepLink} in group ${build.logs.groupName} and stream ${build.logs.streamName}`)
```
katstevens, line 89: Typo - `availabile`

ci/scripts/run-codebuild/libs/codebuild-runner.js
```diff
@@ -0,0 +1,158 @@
...
+    while (true) {
+      for (const logEvent of response.events) {
+        // The build logs already have newlines on the end
+        process.stdout.write(logEvent.message)
+      }
+
+      // When the nextForwardToken is the same as the nextToken we passed in
+      // we have reached the end of the stream. This is different than every other
+      // API in the aws-sdk annoyingly where it would return undefined/null as the next
+      // token.
```
katstevens, line 144: 🤯

ci/scripts/run-codebuild/libs/codebuild-runner.js
```diff
@@ -0,0 +1,158 @@
...
+      if (response.nextForwardToken === nextToken) {
+        return response.nextForwardToken
+      }
+
+      nextToken = response.nextForwardToken
+      command = new GetLogEventsCommand({
+        ...getLogEventsConfig,
```
katstevens, line 151: What are these three dots doing?

katstevens (commented):
First batch of comments - haven't looked at the tests yet

.github/workflows/test-run-codebuild.yml
jfharden: Doh :( Fixed :)

ci/scripts/run-codebuild/libs/codebuild-runner.js
jfharden: It takes getLogEventsConfig and expands it so every attribte in that becomes an attribute in this, then the nextToken is overridden on the next line. It's like saying "merge everything below on top of getLogEventsConfig"

For a better more formal definition see the MDN docs for javascript spread (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)

.github/workflows/test-run-codebuild.yml
jfharden: There we go :) Tested on 12, 14, and 16 :) Passing in all with no changes

jfharden: I've allowed 14 and 16 to float up within the major version, since this is to run on node-runner and currently that's on 12.22.7 it makes sense to lock to the specific target, but where we don't have a specific target yet we may as well float to stay up to date within the LTS release.

ci/scripts/run-codebuild/package.json
jfharden: I could just do >=12.22.7, I've tested it all the way to node17 and everything works without problem so we don't currently have an upper bound

ci/scripts/run-codebuild/libs/codebuild-runner.js
jfharden: Ah yeah. That's weird, I don't know why but my spellcheck sees this as a valid word O_o It's clearly not! I wonder if I accidentally added it to a custom wordlist 🤔

jfharden: Sorted 👍

.github/workflows/test-run-codebuild.yml
```diff
@@ -0,0 +1,42 @@
...
+    name: Test script run-codebuild
+    env:
+      working-directory: ci/scripts/run-codebuild
+
+    strategy:
+      matrix:
+        node-version:
+          - "12.22.7"
+          - "14"
+          - "16"
```
oswaldquek, line 25: Why three node versions? And will the steps below be run three times as a result?

ci/scripts/run-codebuild/README.md
```diff
@@ -0,0 +1,39 @@
...
+}
+```
+
+The fields in this are:
+
+field | description
+---|----
+projectName | The AWS CodeBuild project name to run
+sourceVersion | The source version for the primary source of the CodeBuild project
+secondarySourcesVersions | A map with keys for the secondary source names and values as the versions to use for those sources
```
oswaldquek, line 38: I'm not sure why `secondarySourcesVersions` is needed. Is it related to `sourceVersion`?

oswaldquek: Is `sourceVersion` the source version of the CodeBuild project?
From the example config it looks to me like `secondarySourcesVersions` is referencing some frontend version but there's already `PAY_FRONTEND_RELEASE` which leaves me confused.

ci/scripts/run-codebuild/README.md
```diff
@@ -0,0 +1,39 @@
...
+environmentVariables | A map of with keys as environment variable names to set and values as the values of those environment variables
```
oswaldquek, line 39: "A map with keys"

jfharden: It should be read as "A map, with key as environment variables" I will think how to rephrase this so it's clearer what I mean and update it later 🤔

.github/workflows/test-run-codebuild.yml
jfharden: We are testing for future upgrades (given we know we are moving shortly to either 14 or 16). The tests below do run 3 times as a result, once on each node version

ci/scripts/run-codebuild/README.md
jfharden: Codebuild builds take multiple sources.

It's easier to conceptualise it as github repos rather than files on s3 so lets think of it in that regard for a moment.

A build could have say pay-frontend as it's source (also referred to as the primary source in some of the aws docs). But also have pay-ci and pay-infra as secondary sources.

When you execute the build you specify the sourceVersion as the github ref for the version you want to pull of pay-frontend. You specify refs for both pay-ci and pay-infra in secondarySourcesVersions

In the case of the end to end tests we only have a single source, and the source is from S3, this means we specify the version ID of the file in S3 in these (which was obtained by the upload

ci/scripts/run-codebuild/README.md
jfharden: I just realised as well map is not the right word for json, I guess it's a json object 🤔

ci/scripts/run-codebuild/README.md
oswaldquek: > When you execute the build you specify the sourceVersion as the github ref for the version you want to pull of pay-frontend.

But I'm just seeing
```
"sourceVersion": "123456"
```
How do I infer it's a version for pay-frontend in this case?

oswaldquek: > In the case of the end to end tests we only have a single source, and the source is from S3, this means we specify the version ID of the file in S3

Presumably this source is the pay-ci repo?

ci/scripts/run-codebuild/libs/codebuild-runner.test.js
```diff
@@ -0,0 +1,250 @@
...
+    }
+  ]
+}
+
+describe('CodeBuildRunner', () => {
+  beforeEach(() => {
+    process.stdout.write = jest.fn()
+    cbMock.reset()
+    cwlMock.reset()
+    cwlMock.onAnyCommand().rejects('CLOUDWATCH CALL')
```
katstevens, line 69: Checking my understanding: does this line stub out everything by default, then below we return responses for specific commands?

ci/scripts/run-codebuild/README.md
jfharden: The source is the zip of the pay-ci file on s3. So the sourceversion is the version id of that file.

The run-codebuild task is generic to run any codebuild, so you have to know what the codebuild you are executing has configured as it's source to know what the sourceVersion refers to. In the case of secondary sources they are explicitly named.

In the tests the 123456 is just some fake data to test that what gets sent to codebuild is the same as what was specified by the request.

ci/scripts/run-codebuild/libs/codebuild-runner.test.js
```diff
@@ -0,0 +1,250 @@
...
+      ],
+      [
+        { name: 'MOCK_ENV_VAR_1', value: 'MOCK_ENV_VAR_1_VALUE', type: 'PLAINTEXT' },
+        { name: 'MOCK_ENV_VAR_2', value: 'MOCK_ENV_VAR_2_VALUE', type: 'PLAINTEXT' }
+      ],
+      'eu-west-1',
+      0
+    )
+
+    await expect(runner.runBuildAndLog()).rejects.toThrow('mocked rejection')
```
katstevens, line 224: Is it worth asserting the `process.stdout.write` calls for these error cases like we do for the success case? ('No' is an acceptable answer here!)

ci/scripts/run-codebuild/libs/executor.test.js
```diff
@@ -0,0 +1,86 @@
...
+  },
+  startTime: Date.parse('2021-07-05T10:00:00Z'),
+  endTime: Date.parse('2021-07-05T10:10:00Z')
+}
+
+jest.mock('./codebuild-runner.js', () => {
+  return {
+    CodeBuildRunner: jest.fn().mockImplementation(() => {
+      return {
+        getBuildInfo: jest.fn().mockReturnValue(mockBuildInfo),
```
katstevens, line 26: Are we testing the error case where this is used?

katstevens (commented):
Comments on tests

ci/scripts/run-codebuild/libs/codebuild-runner.test.js
oswaldquek, line 224: How is it throwing a string here and not an `new Error`?

ci/scripts/run-codebuild/README.md
oswaldquek: > The run-codebuild task is generic to run any codebuild, so you have to know what the codebuild you are executing has configured as it's source to know what the sourceVersion refers to.

I understand. I think it's potentially confusing but it's one of things where I'll have to see its use within concourse to get a better handle of it. I'm not suggesting any code changes here.

ci/scripts/run-codebuild/libs/codebuild-runner.test.js
jfharden: The toThrow matcher when provided with a string is looking for an Error with that string as the message, so it is :)

ci/scripts/run-codebuild/libs/executor.test.js
jfharden: I'm not, there was so much going on I decided to leave this case out to be honest

katstevens: Fair enough - we can always revisit in the future

ci/scripts/run-codebuild/libs/codebuild-runner.test.js
jfharden: That's correct

jfharden: Well sort of correct. It's a mock already which I believe will return undefined for anything not actively stubbed. So I'm adding this to make sure if there any unexpected calls to cloudwatch it will error

ci/scripts/run-codebuild/libs/codebuild-runner.test.js
```diff
@@ -0,0 +1,250 @@
...
+  afterEach(() => {
+    process.stdout.write.mockRestore()
+  })
+
+  it('Runs a build, waits for success, and logs the results', async () => {
+    expect.assertions(9)
+
+    let batchGetBuildsCallCount = 0
+
+    cbMock.onAnyCommand().rejects('Unkown CodeBuild API Call')
```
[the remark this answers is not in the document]
jfharden: I feel quite strongly that my vim spelling might not be on again 🤔

ci/scripts/run-codebuild/libs/codebuild-runner.test.js
jfharden: Added :) Thanks! Great spot!

ci/scripts/run-codebuild/package.json
jfharden: Done :)

ci/scripts/run-codebuild/README.md
jfharden: @oswaldquek I updated these descriptions completely, would you mind checking them again?

ci/scripts/run-codebuild/libs/codebuild-runner.test.js
oswaldquek: TIL

oswaldquek (commented):
LGTM, but should get an approval from @katstevens too.

katstevens (approved):

PR #593, opened 2022-02-22
PP-9305: sync non-govukpay dockerhub images

jfharden:
Sync images from dockerhub to ecr.

I did this so there is a reusable task to sync arbitrary images which we can reuse if future if required. The task takes a command separated list of images with tags to copy into ECR.

This can be seen working here: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/e2e-helpers/jobs/sync-images-from-dockerhub-to-ecr/builds/13


katstevens:
Is there a reason we're not using resources as inputs (like we do for the existing [pull-image-from-dockerhub](https://github.com/alphagov/pay-ci/blob/master/ci/tasks/pull-image-from-dockerhub.yml) task)? While this solution feels simpler within the task itself, the visualisation on the Concourse pipeline is a bit harder to see what's coming in and going out.

jfharden:
Yes, it will query dockerhub every minute for new versions for every resource if we do that, we can change that, but we can't limit it to once a week for example, the intervals only stretch to a number of hours, and we can't control when it checks either, it's only an interval, not a date/time :(

jfharden:
You can set never, but then it wont check when a new job starts either, you have to manually click it in the UI or trigger it via an API call (which you can't do within the job since it's too late once the job has started to get new input versions :()

jfharden:
We _could_ break it out into 3 separate tasks instead of 1 task so the tasks can be called copy-<image> but then the yml of the pipeline will expand a lot unless we enable the experimental `across step` (which I would really like, but we already have 1 experimental feature :( and I don't like the idea of using even more)

jfharden:
We could set ecr resources to be never though since we aren't querying it for versions.

There is an advantage to doing this approach, in that we pick up new images within <interval> time of them becoming available (we could set say 1, 6, or 24 hours for example)

katstevens:
> Yes, it will query dockerhub every minute for new versions for every resource if we do that, we can change that, but we can't limit it to once a week for example, the intervals only stretch to a number of hours, and we can't control when it checks either, it's only an interval, not a date/time :(

Aha, that makes sense. Would it be a lot of faff to split out the push-to-ECR step into a separate task so that we get the 3 resource outputs showing? Basically I am keen on being able to click the e2e-helpers pipeline and see at-a-glance what those helpers are. (This shouldn't be a blocker though if it does turn out to be a faff.)

jfharden:
It's definitely a faff, I could still do it though if you would really prefer that, likely take the rest of the day since it means new tasks, saving the docker image and importing it again, setting up the input output chains etc. It also makes this method of doing it pretty pointless since it's not easy to add new ones anymore, and the task can't be generic because you have to define the resources in the pipeline not in the task (I _really_ wish templated pipelines were a thing and not just tasks 😭 )

If we do it as 3 separate then it makes more sense to change this to use the existing pull from dockerhub task, write a new task for push to ecr and make it 2 lots of 3 tasks running in parallel.

So yeah it's a faff, but it would be better in the visualisation for sure.

katstevens:
How about this for a compromise - could we rename the job `sync-generic-images-from-dockerhub-to-ecr`? Then it's a bit clearer which images are being dealt with in that box?

jfharden:
Sure :)

jfharden:
After a chat we agreed to change it to 3 resources in and 3 resources out but with the dockerhub image polling set to every 1 hour

jfharden:
I've changed it to using image resources with a job for each image.
Screenshot 2022-02-22 at 12 04 23

I agree this is better overall for future-us :)

katstevens (approved):
Looking good - visualisation on the pipeline is very clear now 👍

jfharden:
Pipeline fully applied

PR #596, opened 2022-02-22
Bau update runners

jfharden:
Update concourse-runner:
1. Bump to latest docker source container (the current one has multiple critical vulns)
2. Remove the things needed to build appmetrics
3. Compress 3 RUN layers into a single layer, the third layer does an rm of the file retrieved in the first RUN, this does mean it's not in the final layer of the image but it's still part of the container at a lower layer so it doesn't reduce the size as 3 layers.
4. We don't need to get some deps from the edge repo anymore since we moved to a later docker container build so just source them from the primary apk repo.

Update node-runner:
1. Bump to node 12.22.10 and alpine 3.15

# How?

## node-runner
```
$ cd ci/docker/node-runner
$ docker build -t node-runner:test .
```

## concourse-runner
```
$ cd ci/docker/concourse-runner
$ docker build -t concourse-runner:test .
```

### DIFF
--- a/ci/docker/concourse-runner/Dockerfile
+++ b/ci/docker/concourse-runner/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker:19.03
+FROM docker:20.10.10
 
 RUN apk add --no-cache \
   aws-cli \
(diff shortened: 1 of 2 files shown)


ci/docker/concourse-runner/Dockerfile
```diff
@@ -10,26 +10,19 @@ RUN apk add --no-cache \
...
-  # dockerd dependencies
-  iptables \
-  util-linux 
-
-RUN apk add --no-cache \
   openjdk11 \
   npm \
   maven \
   tini \
-  --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community
```
katstevens, line 22: Ah glad we can move off this 👍

ci/docker/concourse-runner/Dockerfile
```diff
@@ -10,26 +10,19 @@ RUN apk add --no-cache \
...
-
-RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub
-RUN wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.31-r0/glibc-2.31-r0.apk
-RUN apk add glibc-2.31-r0.apk
```
katstevens, line 26: 🔥

katstevens (approved):
Great stuff

ci/docker/concourse-runner/Dockerfile
oswaldquek: How do we know all these could be deleted?

PR #597, opened 2022-02-22
PP-9229: Add XXX-candidate tag when pushing image to ECR

oswaldquek:
Add XXX-candidate for a future job (run-frontend-e2e) to be triggered on.
According to
https://github.com/concourse/registry-image-resource#out-push-and-tag-an-image:
```
With additional_tags given in params, the image will be pushed as each tag
listed in the file (whitespace separated).
```


jfharden (approved):
Nice simple change! Very cool! 👍

PR #599, opened 2022-02-23
PP-9229 run frontend e2e task

oswaldquek:


ci/pipelines/deploy-to-test.yml
```diff
@@ -977,6 +985,50 @@ jobs:
         params:
           image: image/image.tar
           additional_tags: tags/tags
+
+  - name: run-frontend-e2e
+    plan:
+      - get: frontend-candidate-ecr-registry-test
```
jfharden, line 991: We could put the 2 gets (this and pay-ci) together in_parallel which speeds up test startup. We don't need to though.

jfharden (approved):
Just one minor thing which we don't _need_ to do, but might be nice

PR #610, opened 2022-02-25
PP-9338: Disable slack notifications for e2e tests

jfharden:
While we are developing the new e2e tests it's proving to be noisy, so disable the notifications for now.


katstevens (approved):

jfharden:
Pipeline set

PR #614, opened 2022-02-25
Pp 9338 run pr e2e tests in codebuild

jfharden:
This PR allows end to end tests to be run on PRs pre-merge.

Changes include:
* Modifying the prepare codebuild task to allow the repository for the app under test to be modified (in order to set it to the pull-request-builds ecr repo)
* Add a task which parses the required information about the build and puts it into files which can be used by an ecr PUT task (and also loaded into the env for later tasks)
* Modify the endtoend-card-e2e job to run in codebuild instead of directly in a concourse container.

Even through YAML anchors are used extensively in this file I've taken the active decision to not use them in this PR. When all of the remaining end to end jobs are changed there will be significantly fewer anchors and the file will be easier to comprehend, if a lot more verbose.

You can see the modified pipeline here: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci?group=end_to_end
You can see a successful test run here: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/endtoend-card-e2e/builds/116

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -506,8 +506,7 @@ resources:
     icon: github
     source:
       uri: https://github.com/alphagov/pay-ci
-      # branch: master
-      branch: pp-9338-run-pr-e2e-tests-in-codebuild
+      branch: master
       username: alphagov-pay-ci
       password: ((github-access-token))
 
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -884,44 +884,82 @@ jobs:
       - <<: *build-docker-image
         params:
           app_name: endtoend
-      - task: get-docker-image-info
-        config:
-          platform: linux
-          image_resource:
-            type: registry-image
-            source:
-              repository: govukpay/concourse-runner
-          inputs:
-            - name: src
-          outputs:
-            - name: image_info
-          run:
-            path: /bin/bash
-            args:
-              - -ec
-              - |
-                mkdir -p image_info/
+      - in_parallel:
+        - task: get-docker-image-info
+          config:
+            platform: linux
+            image_resource:
+              type: registry-image
+              source:
+                repository: govukpay/concourse-runner
+            inputs:
+              - name: src
+            outputs:
+              - name: image_info
+            run:
+              path: /bin/bash
+              args:
+                - -ec
+                - |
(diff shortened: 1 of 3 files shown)


ci/pipelines/pr.yml
```diff
@@ -492,7 +506,8 @@ resources:
     icon: github
     source:
       uri: https://github.com/alphagov/pay-ci
-      branch: master
+      # branch: master
+      branch: pp-9338-run-pr-e2e-tests-in-codebuild
```
oswaldquek, line 510: Should this be `master`?

oswaldquek (changes requested):
Largely LGTM!

ci/pipelines/pr.yml
jfharden: Indeed it should 🤦 I didn't mean to include this in my commit! Doh!

jfharden: Fixed

oswaldquek (approved):

ci/tasks/get-pr-build-docker-image-info.yml
```diff
@@ -0,0 +1,43 @@
...
+  args:
+    - -ec
+    - |
+      mkdir -p image_info/
+
+      pr_number=$(cat src/.git/resource/pr)
+      head_sha=$(cat src/.git/resource/head_sha)
+
+      # We cannot access the concourse build number in a task but we need a unique
+      # tag for our build so lets just use a UUIDv4
```
katstevens, line 38: That's annoying - I guess we'd have to pass through `$BUILD_NAME` as an input if we wanted to do this? UUID is fine here though

katstevens (approved):

ci/tasks/get-pr-build-docker-image-info.yml
jfharden: Sadly the env vars are not available in the pipeline definition either, so we would have to write them out to a file which could be consumed by load var, or the task :( It's a conscious decision on the part of the concourse designers. Someone wrote a resource which would do exactly write it out and load it into a var for you, but I didn't think it was worth bringing in another concourse resource type just to avoid using a random id.

PR #616, opened 2022-02-28
PP-9338: Change card-connector pr e2e tests to use codebuild

jfharden:
Change connector, publicapi, and adminusers to use end to end tests in codebuild.


katstevens (approved):
Good spot on the products-ui stuff!

PR #617, opened 2022-02-28
Pp 9338 change products UI e2e tests to codebuild

jfharden:
This PR changes the final set of e2e tests, I had no choice but to remove the unused resources since the pre-commit checks fail with a pipeline that has unused resources.

With products ui the project name is products-ui but env vars can't have hyphens, and in the docker-compose files they are already listed with the hyphens so I amended the prepare-codebuild task to remove any hyphens for use in env var names.

A working build can be seen in https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/products-ui-e2e/builds/4

If you expand the task `prepare-codebuild` you can see the env var renaming was successful:

Screenshot 2022-02-28 at 15 39 33

Additionally if you look in the run-products-e2e-tests output after the pull is complete you can see the products-ui container running with the correct image:

Screenshot 2022-02-28 at 15 43 07

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -115,39 +115,6 @@ definitions:
     passed: [updateThisValue]
     params:
       filename: image-updateThisValue-*.tar
-  - &put-card-e2e-pending-status
-    put: updateThisValue
-    get_params:
-      skip_download: true
-    params:
-      path: src
-      status: pending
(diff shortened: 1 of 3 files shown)


katstevens (approved):
🏆  Loving the diff on this one

PR #621, opened 2022-03-01
Pp 9303 add reusable GitHub workflows

jfharden:
Add 2 simple reusable workflows we can call from other repos (or this repo).

One to lint and test node apps

One to create an alpha release tag.

You can see the unit test one in action in this pr: https://github.com/alphagov/pay-stubs/pull/354 and the checks it ran: https://github.com/alphagov/pay-stubs/actions/runs/1917470455


.github/workflows/_create-alpha-release-tag.yml
```diff
@@ -0,0 +1,37 @@
...
+permissions:
+  contents: read
+
+jobs:
+  tag-release:
+    needs: unit-tests
+    runs-on: ubuntu-20.04
+
+    permissions:
+      contents: write
```
katstevens, line 15: Given this has repo write permissions, we need to be a little careful when we invoke this from elsewhere. [Github recommends pinning to a ref](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow) - `@master` is probably fine - changes to this file will have been code-reviewed etc. I don't think we can enforce this here, but perhaps we could add a comment with an example of expected usage? Something like

```
# Usage: 
# jobs:
#  call-create-alpha-release-from-pay-ci:
#    uses: alphagov/pay-ci/.github/workflows/_create-alpha-release-tag.yml@master
```

jfharden: Sure, adding some help text is fine, although as I understand it pinning to @master is the same as not pinning, except in the case that the primary branch name is changed.

jfharden: Just an aside again, about how much I dislike the github permissions model :( I really wish it were more granular so you could grant "make a release" type permissions without requiring access to the whole repo :(

jfharden: Added to the top of the workflow (I've actually added help info to both since it's a good idea! :) )

PR #622, opened 2022-03-02
PP-9303: Allow a release tag prefix in the reusable release workflow

jfharden:
Allow the create release reusable workflow to accept an additional prefix to add to the tag. This is useful in monorepos which have multiple artefacts to release.

### DIFF
--- a/.github/workflows/_create-alpha-release-tag.yml
+++ b/.github/workflows/_create-alpha-release-tag.yml
@@ -14,6 +14,12 @@ name: Create alpha_release tag
 
 on:
   workflow_call:
+    inputs:
+      tag_prefix:
+        description: An additional prefix to add the to the tag. This must not contain any hyphens!
+        default: ""
+        required: false
+        type: string
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/_create-alpha-release-tag.yml
```diff
@@ -33,13 +39,13 @@ jobs:
           fetch-depth: '0'
       - name: Tag release
         run: |
-          LATEST_RELEASE_NUMBER=$(git describe --abbrev=0 --tags --match "alpha_release-*" | awk -F- '{print $2}' || true)
+          LATEST_RELEASE_NUMBER=$(git describe --abbrev=0 --tags --match "${{ inputs.tag_prefix }}alpha_release-*" | awk -F- '{print $2}' || true)
```
oswaldquek, line 42: What might be a literal value of `${{ inputs.tag_prefix }}` here? Looking at https://payments-platform.atlassian.net/browse/PP-9303, what repos would this apply to (i.e. stubs, endtoend, pay-scripts)?

jfharden: for pay-scripts `proxy_` so it would become `proxy_alpha_release-XXX`

jfharden: Right now, pay-dockerfiles, pay-scripts, pay-ci potentially

PR #629, opened 2022-03-03
PP-9361: Run pr e2e tests automatically 

jfharden:
Run all PR e2e tests automatically and without waiting for unit tests to pass first

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -605,8 +605,7 @@ jobs:
     plan:
       - <<: *get-pull-request
         resource: card-connector-pull-request
-        trigger: false
-        passed: [card-connector-unit-test]
+        trigger: true
       - in_parallel:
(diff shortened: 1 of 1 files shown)


ci/pipelines/pr.yml
```diff
@@ -605,8 +605,7 @@ jobs:
     plan:
       - <<: *get-pull-request
         resource: card-connector-pull-request
-        trigger: false
-        passed: [card-connector-unit-test]
+        trigger: true
```
katstevens, line 608: We should probably be consistent and either have `trigger: true` explicitly written out for all the tasks (not just here), or remove this one?

jfharden: Removed 👍

katstevens (approved):
Woohoo let's gooooo 💨

jfharden:
They pass on a new pr on master so I'm happy

PR #632, opened 2022-03-03
[PP-9344] Replace "Abandonware" SQS Image

dj-maisy:
## What?
This swaps the `roribio16/alpine-sqs` image with a drop-in replacement from `softwaremill/elasticmq` to represent the SQS service during testing.

## Why?
The existing `roribio16/alpine-sqs` image had not been updated in a very long time, so we should replace it with an up-to-date solution where possible.


katstevens:
@dj-maisy could you add in a link to the source/changelog of this new image in the PR, so we have a reference if we want to update it later? 

While this is only to be used in e2e tests, it will be potentially running on developer machines so also good to add a note about any security considerations (e.g stating that the replacement image is from a reputable/well-documented/maintained source).

jfharden:
You'll need to locally change the branch the pipeline uses of ci to use this branch so it uses your new docker-compose files though

dj-maisy:
Running locally:
```
docker-compose up --force-recreate -d
docker ps -aqf "name=endtoend"
docker exec a64177b8ff2c /app/bin/e2e-card
```

Results:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  14:45 min
[INFO] Finished at: 2022-03-04T15:54:16Z
[INFO] ------------------------------------------------------------------------
```

katstevens (approved):

PR #633, opened 2022-03-04
PP-9234: Require e2e tests for adminusers

jfharden:
This will:

1. Rename `push-adminusers-to-test-ecr` to `push-adminusers-candidate-to-test-ecr`
2. Push X-release to test ecr only _after_ e2e tests are complete
3. Require e2e tests before running deploy-adminusers
4. Push latest-master to dockerhub after e2e-tests
5. Rename the canididate and latest ecr resources to just `adminusers-(latest|candidate)` I didn't rename the release since that would lose the deploy history and could trigger a redeploy of lots of old versions

New pipeline layout:
Screenshot 2022-03-04 at 09 37 21

https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test?group=adminusers

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -279,14 +279,14 @@ resources:
       repository: govukpay/adminusers
       variant: release
       <<: *aws_test_config
-  - name: adminusers-candidate-ecr-registry-test
+  - name: adminusers-candidate
     type: registry-image
     icon: docker
     source:
       repository: govukpay/adminusers
       variant: candidate
       <<: *aws_test_config
-  - name: adminusers-latest-ecr-registry-test
+  - name: adminusers-latest
     type: registry-image
     icon: docker
     source:
@@ -1385,7 +1385,7 @@ jobs:
         file: pay-ci/ci/tasks/parse-release-tag.yml
         input_mapping:
           git-release: adminusers-git-release
-      - put: adminusers-candidate-ecr-registry-test
+      - put: adminusers-candidate
         params:
           image: image/image.tar
           additional_tags: tags/candidate-tag
@@ -1393,7 +1393,7 @@ jobs:
   - name: run-adminusers-e2e
     plan:
       - in_parallel:
-        - get: adminusers-candidate-ecr-registry-test
+        - get: adminusers-candidate
           params:
             format: oci
           trigger: true
@@ -1403,9 +1403,9 @@ jobs:
         - task: parse-candidate-tag
           file: pay-ci/ci/tasks/parse-candidate-tag.yml
           input_mapping:
-            ecr-repo: adminusers-candidate-ecr-registry-test
+            ecr-repo: adminusers-candidate
         - load_var: candidate_image_tag
-          file: adminusers-candidate-ecr-registry-test/tag
+          file: adminusers-candidate/tag
         - task: assume-role
           file: pay-ci/ci/tasks/assume-role.yml
           params:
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -1422,9 +1429,18 @@ jobs:
...
+      - in_parallel:
+        - do:
+          - put: adminusers-ecr-registry-test
+            params:
+              image: adminusers-candidate/image.tar
+              additional_tags: parse-candidate-tags/release-tag
+          - put: adminusers-latest
+            params:
+              image: adminusers-candidate/image.tar
+        - put: adminusers-dockerhub
```
katstevens, line 1441: Am I right that, until we remove the Jenkins step, this `latest-master` image will already be in Dockerhub?

ci/pipelines/deploy-to-test.yml
```diff
@@ -1422,9 +1429,18 @@ jobs:
...
+          params:
+              image: adminusers-candidate/image.tar
 #    on_failure:
```
katstevens, line 1444: Are we ready to un-comment these notifications yet?

ci/pipelines/deploy-to-test.yml
jfharden: That's correct, but it doesn't really matter, it _should_ be a quick push because it's using the same image so I think all these will say "existing blob"

jfharden: Yeah the one that ran:
Screenshot 2022-03-04 at 10 02 48

ci/pipelines/deploy-to-test.yml
jfharden: Yeah I think we need to but I wanted to test this before I annoyed everyone :D Will do it in this PR now it's tested :)

jfharden: Added 👍 thanks :)

ci/pipelines/deploy-to-test.yml
```diff
@@ -1422,34 +1429,44 @@ jobs:
...
-#        text: ':green-circle: adminusers candidate image ((.:candidate_image_tag)) passed post-merge end to end tests - <https://pay-cd.deploy.payments.service.gov.uk/teams/$BUILD_TEAM_NAME/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME|Concourse build #$BUILD_NAME>'
-#        icon_emoji: ":concourse:"
-#        username: pay-concourse
+      - in_parallel:
+        - do:
+          - put: adminusers-ecr-registry-test
+            params:
+              image: adminusers-candidate/image.tar
+              additional_tags: parse-candidate-tag/release-tag
+          - put: adminusers-latest
```
katstevens, line 1438: Hang on - are we pushing to Dockerhub twice (`latest` here, and `latest-master` below)?

jfharden: No, this is ECR, we are pushing to ecr twice though (it will require a big change to be able to do it in 1 push) but the second push is v quick since it's an already exists, and the two together will take less time than the single push to dockerhub so the flow is

* In parallel
    * in series
        * push X-release
        * push latest
    * push to dockerhub
    
So I still expect the push to ecr tasks to finish prior to the parallel dockerhub

katstevens: Gah... it turns out the line 291 that says `icon: docker` confused me 🤦‍♀️  (that'll serve me right to try and multi-task reviewing big yaml files)

jfharden: Yeah I wish we had an iconset for aws services, would be nice to have an ECR icon :)

katstevens (approved):

PR #645, opened 2022-03-09
PP-9253: Build job for alphagov/pay-stream-s3-sqs

oswaldquek:
Build docker image and push to ECR.

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -121,6 +121,13 @@ resources:
       password: ((github-access-token))
 
   # Github Releases
+  - name: stream-s3-sqs-git-release
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-stream-s3-sqs
+      branch: master
+      tag_regex: "alpha_release-(.*)"
   - name: toolbox-git-release
     type: git
     icon: github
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -709,6 +723,9 @@ resource_types:
       tag: latest
 
 groups:
+  - name: stream-s3-sqs
```
jfharden, line 726: Please can you move this to later in the list of groups, this would make it the first thing in the headings in the deploy-to-test pipeline (since the UI is ordered the same as this groups list).

Currently the order is:

1. all apps alphabetically
2. all supporting things mostly-alphabetically (notifications is in the wrong place sitting between telegraf and update-deploy-to-test-pipeline)

So if you could move this to between nginx-forward-proxy and telegraf that would be good 👍

jfharden (commented):
Just a small thing.

Could you set this pipeline after the small change and make sure it works?

ci/pipelines/deploy-to-test.yml
oswaldquek: [Done](https://github.com/alphagov/pay-ci/pull/645/commits/fb05adf865e241139ced0650d9bc0bbd75dcf93b). Looks like it works: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/build-and-push-stream-s3-sqs-to-test-ecr/builds/1.

jfharden (approved):
Crackin' 👍

PR #647, opened 2022-03-15
PP-9236: Workflow for running a provider contract test

oswaldquek:
The pro of having this workflow is that the maven command to run this test is
the same across all our providers.

The cons:
* The providers must be in Java
* The provider project must have a pom.xml which defines a `runContractTests` profile

[adminusers PR](https://github.com/alphagov/pay-adminusers/pull/1287) that proves the use of this workflow.

I've tested that we don't need to pass `-Dpact.provider.version=${commit}` to
the `mvn test -DrunContractTests` command as the pact library will pull out the
latest sha of the provider when publishing the result.

### DIFF
--- a/.github/workflows/_run-provider-contract-tests.yml
+++ b/.github/workflows/_run-provider-contract-tests.yml
@@ -1,0 +1,43 @@
+# This workflow is meant to be run on a pact provider build. It will run
+# all the consumers that depend on it against itself.
+#
+name: Provider Contract Tests
+
+on:
+  workflow_call:
+    secrets:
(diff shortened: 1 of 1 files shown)


jfharden (approved):
We can always expand this (or have an alternate) if we have any non-java providers in future. E.g. 2 steps, 1 for java, 1 for node and which one runs is picked by a param, or just a second workflow.

For now though this is great and a big improvement on the same code in lots of repos :)

katstevens (approved):

jfharden (approved):

PR #651, opened 2022-03-21
PP-9237 Deploy control plane in Concourse

katstevens:
https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/control-plane

Adds a build-&-deploy job to the existing Control Plane pipeline:

- new resource: control plane git repo (to checkout and build the image from)
- new resource: control plane dockerhub repo (to be pushed to following the build)
- task: build control plane using `concourse/oci-build-task`
- task: push to dockerhub using tag `latest` (Jenkins currently pushes `latest-master` so this won't conflict)
- task: push to PaaS using the existing cf resource already used by the restart job
- success/failure slack notifications 

At present Jenkins pushes the `latest-master` image to Dockerhub, then applies the `cf push` calling the image _from Dockerhub_, not locally. I've done the same here for now as it's worked out easier with the concourse cf resource. Our other PaaS apps aren't deployed as docker images (the concourse cf resource uses the manifest instead).

~Trigger currently set to false while testing.~ The pipeline will now trigger the deploy on a merge to the Control Plane repo (as I suspect a few devs mistakenly believe Jenkins currently does...)

### DIFF
--- a/ci/pipelines/control-plane.yml
+++ b/ci/pipelines/control-plane.yml
@@ -10,6 +10,24 @@ resources:
       username: ((cf-username))
       password: ((cf-password))
 
+  - name: pay-control-plane-src
+    type: git
+    icon: github-circle
+    source:
+      uri: https://github.com/alphagov/pay-control-plane
+      branch: master
+      username: alphagov-pay-ci
+      password: ((github-access-token))
+
+  - name: pay-control-plane-latest
+    type: registry-image
+    icon: docker
+    source:
+      repository: govukpay/control-plane
+      username: ((docker-username))
+      password: ((docker-password))
+      tag: latest
+
   - name: every-morning
     type: time
     icon: alarm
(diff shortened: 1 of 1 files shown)


ci/pipelines/control-plane.yml
```diff
@@ -10,6 +10,24 @@ resources:
       username: ((cf-username))
       password: ((cf-password))
 
+  - name: pay-control-plane-src
+    type: git
+    icon: github-circle
```
jfharden, line 15: think this needs to just be github. It's showing just a question mark atm. It has to be an icon available in https://materialdesignicons.com/

katstevens: Aha - https://github.com/alphagov/pay-team-manual/blob/master/ci/pipeline/pay-team-manual.yml#L11

jfharden:
Hmm a philosophical question.

Should we have github actions create a release tag and deploy from that (as the other apps do)? 🤔

ci/pipelines/control-plane.yml
jfharden: Oh yeah, that one is also broken 😆
Screenshot 2022-03-21 at 16 06 34

ci/pipelines/control-plane.yml
```diff
@@ -62,3 +80,55 @@ jobs:
...
+        inputs:
+          - name: pay-control-plane-src
+            path: .
+        outputs:
+          - name: image
+        run:
+          path: build
+    - put: pay-control-plane-latest
+      params:
+        image: pay-control-plane-image/image.tar
```
jfharden, line 108:
```suggestion
        image: pay-control-plane-image/image.tar
        additional_tags: pay-control-plane-src/.git/ref
```

In defence of this: pay-control-plane is on a knife edge barely working and only just buildable.....so.....in order to not lose the older images what do you think to adding the git sha as an additional tag so we get a tag of the git commit sha as well meaning as the latest tag moves dockerhub doesn't discard the older versions of the image?

katstevens: Absolutely - that was on my mind earlier and I completely forgot about it. Thanks!

jfharden: Wicked, that worked a treat :)

jfharden (approved):

PR #660, opened 2022-03-23
PP-9213: products and publicapi builds

jfharden:
Build and push both products and publicapi in concourse


katstevens (approved):
Ready when the others are ready 👍

PR #662, opened 2022-03-23
Pp 9213 cardid build

jfharden:
Build cardid in concourse.
Bonus to make the selfservice gets in parallel

This is slightly different than the other java apps since it has a submodule in a private repo. The release checkout is done with `submodules: none` and then a task is run which actually gets the submodule using https basic auth instead of oath. This code was already written I just changed the image it used from concourse-runner (which is huge) to alpine and did apk add git (as we discovered the other day this is about 2 seconds, instead of around a minute pulling the concourse-runner image).

The whole repo is then copied into cardid-git-release-with-submodules output directory to be passed to later tasks. The later tasks all use this with-submodules version of the repo instead of the plain cardid-git-release repo.

Everything else is the same as the other moves to build in concourse.

I've done the same testing we did on our first one and created an additional build job and e2e test job in order to validate this is ok.

The build job: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/build-and-push-cardid-candidate/builds/1
The e2e test job: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/run-cardid-e2e-starling-test/builds/1

I also pulled the image from ECR and checked it included the cardid-data submodule data and it does.

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -3797,9 +3797,10 @@ jobs:
 
   - name: build-and-push-selfservice-candidate
     plan:
-      - get: pay-ci
-      - get: selfservice-git-release
-        trigger: true
+      - in_parallel:
+        - get: pay-ci
+        - get: selfservice-git-release
+          trigger: true
       - task: parse-release-tag
         file: pay-ci/ci/tasks/parse-release-tag.yml
         input_mapping:
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -741,7 +741,7 @@ groups:
       - adminusers-db-migration
   - name: cardid
     jobs:
-      - push-cardid-candidate-to-test-ecr
+      - build-and-push-cardid-candidate
       - run-cardid-e2e
       - deploy-cardid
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -4090,33 +4090,35 @@ jobs:
...
           inputs:
             - name: cardid-git-release
           params:
             GH_ACCESS_TOKEN: ((github-access-token))
           outputs:
-            - name: image
+            - name: cardid-git-release-with-submodules
           run:
-            path: bash
+            path: ash
```
katstevens, line 4116: Is this alpine shell?

jfharden: Yeah

jfharden: I think in alpine sh and ash are actually the same, but ash is slightly different to the sh you get with ubuntu I believe so I like to call out that it's ash not sh really

katstevens (approved):
Pipeline looks good 👍 

While reviewing this I've realised I should probably do some more reading up on submodule behaviour in GHA workflows...

PR #665, opened 2022-03-25
BAU: Set pact provider version for PRs or Master tests

jfharden:
Set the provider version to either the PR sha, or the master sha

This allows the pact tests to be used during a PR and to set a SHA in the broker that we can actually trace back to a PR.

You can see this in use in https://github.com/alphagov/pay-connector/pull/3555

While we do not need this functionality right this second as part of testing how to do it I think this has become ok to merge and makes it trivial for us to move the other tests later.

### DIFF
--- a/.github/workflows/_run-provider-contract-tests.yml
+++ b/.github/workflows/_run-provider-contract-tests.yml
@@ -43,7 +43,7 @@ jobs:
             # in the pact broker something we can actually trace back to a PR, using the _actual_
             # phantom merge commit sha (github.sha) would make this essentially impossible for us
             PROVIDER_VERSION=${{ github.event.pull_request.head.sha }}
-          elif [[ "${{ github.ref }}" == "refs/heads/master" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/_run-provider-contract-tests.yml
```diff
@@ -35,10 +35,26 @@ jobs:
...
           docker pull govukpay/postgres:11.1
+      - name: Set Pact Provider Version
+        id: set-pact-provider-version
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            # Github actually creates a phantom merge commit on PR actions, but we want to record
+            # in the pact broker something we can actually trace back to a PR, using the _actual_
+            # phantom merge commit sha (github.sha) would make this essentially impossible for us
+            PROVIDER_VERSION=${{ github.event.pull_request.head.sha }}
+          elif [[ "${{ github.ref }}" == "refs/heads/master" ]]; then
```
jfharden, line 46: Maybe for future compatibility this should be master or main

oswaldquek (approved):

PR #666, opened 2022-03-29
PP-9428 Retry Worldpay smoke test failures

katstevens:
Allow up to 10 attempts for the Worldpay post-deploy smoke tests before failing. This should reduce noise in the release slack channel and enable more continuous deployment without developer intervention. 

We will still have visibility on continued failures through the scheduled staging smoke tests.


jfharden (approved):
It's hard to adequately express how much I dislike having to do this :( As a stop-gap it's ok since we are following up through support :( 

Thank you for doing this though 👍

PR #667, opened 2022-03-30
BAU: Allow 3 attempts for flaky card e2e tests

katstevens:
Adds retries for the deploy-to-test, e2e-helpers and pr-ci pipelines.

These retries can be removed once the flaky EPDQ scenario is no longer required.


dj-maisy (approved):
It's sad that we need retries on anything, but we know the tests are a bit flappy and for now we need it! 👍

PR #669, opened 2022-03-31
BAU: Remove stubs tests from PR pipeline

katstevens:
These tests are already run on PRs via Github Actions.


alexbishop1 (approved):
Stubbed that right out

PR #683, opened 2022-05-03
PP-8295: Deploy sidecars to perf

oswaldquek:
There is no need for separate deploy jobs for telegraf, nginx-proxy and
nginx-forward-proxy as these are already automatically triggered via app
deployments.


jfharden:
Even without the need for jobs it would be good to create groups for those sidecars which include the jobs that deploy them like we have in the other deploy pipelines (e.g. https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test?group=nginx-forward-proxy has deploy-frontend among others).

It just means when we have to make a deployment of these things happen we don't have to go searching for "which app(s) deploy this"

ci/pipelines/deploy-to-perf.yml
```diff
@@ -832,3 +869,187 @@ jobs:
...
+                terraform init
+                terraform apply \
+                  -var notifications_image_tag=${APPLICATION_IMAGE_TAG} \
+                  -auto-approve
+      - task: wait-for-deploy
+        file: pay-ci/ci/tasks/wait-for-deploy.yml
+        params:
+          APP_NAME: notifications
+          APPLICATION_IMAGE_TAG: ((.:application_image_tag))
+          ENVIRONMENT: staging-2
```
jfharden, line 1038: This needs to be test-perf-1.

jfharden:
@oswaldquek It doesn't need new jobs, just a few groups listing the already existing jobs

oswaldquek:
> @oswaldquek It doesn't need new jobs, just a few groups listing the already existing jobs

@jfharden [done](https://github.com/alphagov/pay-ci/pull/683/commits/f97597a8e4013178a96e23a8a81aa37952ef8194)

jfharden (approved):

jfharden:
Awesome :)

PR #684, opened 2022-05-05
PP-9406 Run cardid provider pact tests on PR builds

stephencdaly:
- Run cardid provider pact tests on connector PR builds and cardid PR builds
- When we checkout cardid master for the connector-as-consumer pact
tests, don't fetch the cardid-data submodule as this isn't required for
pact tests to run.

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -244,6 +244,7 @@ groups:
     jobs:
       - cardid-unit-test
       - cardid-integration-test
+      - cardid-as-provider-pact-test
       - cardid-e2e
       - record-cardid-build-time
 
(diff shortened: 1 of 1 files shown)


jfharden (approved):
Looks great :)

Is connector the only consumer of cardid? If not it might be worth adding the others in too (I thought frontend used it to, but I will admit that's just built on ignorance and assumptions :) )

oswaldquek (approved):
LGTM!

stephencdaly:
> Looks great :)
> 
> Is connector the only consumer of cardid? If not it might be worth adding the others in too (I thought frontend used it to, but I will admit that's just built on ignorance and assumptions :) )

There's currently no contract tests between frontend and cardid. We do have end-to-end tests covering the interaction, so there probably hasn't been a need.

jfharden:
> > Looks great :)
> > Is connector the only consumer of cardid? If not it might be worth adding the others in too (I thought frontend used it to, but I will admit that's just built on ignorance and assumptions :) )
> 
> There's currently no contract tests between frontend and cardid. We do have end-to-end tests covering the interaction, so there probably hasn't been a need.

Thanks for the explanation :)

PR #685, opened 2022-05-05
PP-9266: Wait for migration to complete

oswaldquek:
Example successful build: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/adminusers-db-migration/builds/8

### DIFF
--- a/ci/scripts/run-ecs-db-migration.js
+++ b/ci/scripts/run-ecs-db-migration.js
@@ -118,8 +118,14 @@ const run = async function run () {
     await waitFor(`${runResult.tasks[0].taskArn}`)
 
     const describeTasksResult = await describeTasks(`${runResult.tasks[0].taskArn}`)
-    if (describeTasksResult.tasks[0].containers.some(container => container.exitCode !== 0)) {
+    const containers = describeTasksResult.tasks[0].containers
+    if (containers.some(container => container.exitCode !== 0)) {
       console.log(`One or more containers in the task ${runResult.tasks[0].taskArn} did not exit with 0`)
+      containers.forEach(container => { 
+        if (container.exitCode !== 0) {
+          console.log(`${container.taskArn} exited with ${container.reason}`)
+        }
+      })
       process.exit(1)
     }
 
--- a/ci/scripts/run-ecs-db-migration.js
+++ b/ci/scripts/run-ecs-db-migration.js
@@ -51,12 +51,26 @@ async function runTask (params) {
   }
 }
 
-async function waitFor (params) {
+async function waitFor (taskArn) {
   console.log(`Waiting for database migration to complete...`)
   try {
-    return await ecs.waitFor('tasksStopped', params).promise()
+    return await ecs.waitFor('tasksStopped', {
+      cluster: CLUSTER_NAME,
+      tasks: [taskArn]
+    }).promise()
   } catch (err) {
     throw new Error(`Error attempting to wait for ecs task: ${err.message}`)
+  }
+}
+
+async function describeTasks(taskArn) {
+  try {
+    return await ecs.describeTasks({
+      cluster: CLUSTER_NAME,
+      tasks: [taskArn]
+    }).promise()
+  } catch (err) {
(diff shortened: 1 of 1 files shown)


ci/scripts/run-ecs-db-migration.js
```diff
@@ -90,7 +99,13 @@ const run = async function run () {
     }
 
     console.log(`Succesfully scheduled db migration task: ${runResult.tasks[0].taskArn}`)
-    console.log('View progress of the migration at: https://gds.splunkcloud.com/en-GB/app/gds-004-pay/migration_status')
```
jfharden, line 93: I'd leave this log in at the start, if a migration takes a while the dev can go watch the logs in splunk at least to see what's happening

ci/scripts/run-ecs-db-migration.js
```diff
@@ -90,7 +99,13 @@ const run = async function run () {
...
+
+    await waitFor({
+      cluster: CLUSTER_NAME,
+      tasks: [`${runResult.tasks[0].taskArn}`]
+    })
+
```
jfharden, line 107: I think after the task is stopped you should query what the exit code of the task is, if it's a failure state exit out of this with a failure. Tasks that fail also go into the STOPPED state so a failed migration won't throw any exception.

jfharden (commented):
A small dev experience change, but also suggesting a more important change to make failed migrations cause the task to fail

ci/scripts/run-ecs-db-migration.js
oswaldquek: What about failing the job if the status of the task is anything other than "RUNNING"?

oswaldquek: The available states are at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-lifecycle.html

oswaldquek: So something like 
```
    const waitForResult = await waitFor({
      cluster: CLUSTER_NAME,
      tasks: [`${runResult.tasks[0].taskArn}`]
    })

    if (waitForResult.tasks[0].lastStatus !== 'RUNNING') {
      console.log(`Expected task status 'RUNNING' but was ${waitForResult.tasks[0].lastStatus}.`)
      process.exit(1)
    }
```

jfharden: The state immediately prior to STOPPED is very likely to be DEPROVISIONING rather than RUNNING though. I think you need to use the describe tasks api (I'm sure there's an aws-sdk equivalent, or possibly it's in the waitForResult too) https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeTasks.html to the the exit code (https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Container.html#ECS-Type-Container-exitCode) for the specific container (I think we actually have multiple containers running in the task (e.g. telegraf, nginx, the app, in some cases an nginx-forward-proxy (maybe only in frontend for that one)) and check it's exitCode is 0.

ci/scripts/run-ecs-db-migration.js
oswaldquek: Fixed

ci/scripts/run-ecs-db-migration.js
oswaldquek: https://github.com/alphagov/pay-ci/pull/685/commits/50da474341146b68d67f1d5240b02cc1927ced48.
[Successful DB migration](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/adminusers-db-migration/builds/10).

oswaldquek:
@jfharden I've updated the PR.

ci/scripts/run-ecs-db-migration.js
oswaldquek: Hmm yeah I think that'll be super painful to test. Because a poorly formatted migration script for example would fail the app integration test to begin with.

PR #693, opened 2022-05-10
PP-9307 add adminusers - ledger pact test pipeline

SandorArpa:
- add Concourse PR pipeline for the new pact test, where adminusers is the consumer and ledger is the provider. deploy pipelines are already set up for pact tagging.


stephencdaly (approved):
LGTM!

PR #694, opened 2022-05-11
PP-9582 manifest and pipeline for `pay-stubs`

nlsteers:
### WHAT
- Added a new manifest to deploy `pay-stubs` to paas
- Added a new pipeline definition to continuously deploy `pay-stubs`

### DIFF
--- a/ci/pay_stubs/pay-stubs_manifest.yml
+++ b/ci/pay_stubs/pay-stubs_manifest.yml
@@ -1,0 +1,23 @@
+applications:
+- name: pay-stubs
+  command:
+  disk_quota: 512M
+  PORT: 8080
+  env:
+    DISABLE_INTERNAL_HTTPS: true
+    SMARTPAY_EXPECTED_PASSWORD: ((smartpay-expected-password))
+    SMARTPAY_EXPECTED_USER:     ((smartpay-expected-user))
+    WORLDPAY_EXPECTED_PASSWORD: ((worldpay-expected-password))
+    WORLDPAY_EXPECTED_USER:     ((worldpay-expected-user))
+    CONNECTOR_WORLDPAY_NOTIFICATION_URL: 'https://notifications-test-perf-1.test.pymnt.uk/v1/api/notifications/worldpay'
+    CONNECTOR_SMARTPAY_NOTIFICATION_URL: 'https://notifications-test-perf-1.test.pymnt.uk/v1/api/notifications/smartpay'
+    CONNECTOR_EPDQ_NOTIFICATION_URL: 'https://notifications-test-perf-1.test.pymnt.uk/v1/api/notifications/epdq'
+    EPDQ_SHA_OUT_PASSPHRASE: '' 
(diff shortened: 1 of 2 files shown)


ci/pay_stubs/pay-stubs_manifest.yml
```diff
@@ -0,0 +1,25 @@
...
+- name: pay-stubs
+  command:
+  disk_quota: 512M
+  docker:
+    image: govukpay/stubs
+  env:
+    PORT: 8080
+    DISABLE_INTERNAL_HTTPS: true
+    SMARTPAY_EXPECTED_PASSWORD: ((smartpay-expected-password))
+    SMARTPAY_EXPECTED_USER:     ((smartpay-expected-user))
```
katstevens, line 11: Hopefully we'll be able to remove the Smartpay stuff at some point soon.

ci/pipelines/pay-stubs.yml
```diff
@@ -0,0 +1,74 @@
...
+    icon: cloud-upload-outline
+    source:
+      api: https://api.cloud.service.gov.uk
+      org: govuk-pay
+      space: build
+      username: ((cf-username))
+      password: ((cf-password))
+      cf_cli_version: 7
+
+  - name: stubs-dockerhub
```
katstevens, line 38: Is there a reason we're using Dockerhub rather than the stubs ECR repository? (We build & push stubs to ECR in the [e2e-helpers pipeline](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/e2e-helpers))

katstevens: While it's the same image in Dockerhub, it'll be quicker for Concourse to fetch the image from ECR (no rate limiting etc).

jfharden: Concourse isn't the thing that needs to fetch the image, it's actually PaaS that needs to pull it which doesn't have access to our ECR so I think we have to pull it from dockerhub

jfharden (approved):

nlsteers:
Pipeline in Concourse: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/pay-stubs
`pay-stubs` running on PaaS: https://pay-stubs.cloudapps.digital/

ci/pipelines/pay-stubs.yml
katstevens: Makes sense!

jfharden: Just had a look and it requires an aws access key id and secret access key to do so, which means we would have to supply it with user credentials which feels painful, or orchestrate generating some with STS and passing them (which is much more secure since they would be temporary and we could set a timeout on them of 5 minutes or something).

Either way I think it's worth committing this as-is and changing it on it's own since it would be more complex.

jfharden: See https://docs.cloudfoundry.org/devguide/deploy-apps/push-docker.html#ecr

PR #707, opened 2022-05-27
PP-8337: Add perf-tests pipeline

jfharden:
Add a perf tests pipeline.

Note: While this PR looks huge, there is a lot of repetition because of the nature of these yaml pipelines. I have chosen to make the 3 biggest tasks which get repeated a lot into a YAML anchor, otherwise this yaml file would be something in the region of 10000 lines and an extreme pain to update.

Additional note: This isn't all of the behaviour described in the slack story, but it's a good base to start adding slack notifications etc. I've also chosen to run the tests at a different time than the original story, when the story was written it was predicated on the scaling being done by an external process which ran at fixed times. Moving it to 11pm means it's less likely to interrupt developers trying to actually use the perf test environment.

This gives [a daily job](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/perf-tests?group=complete-perf-tests) which runs at 11pm which:

1. Pauses the deploy to perf pipeline and waits for running jobs to finish
2. Starts the test-perf-1 databases
3. Scales up all the test-perf-1 services to production volumes (including stubs in paas)
4. Loads the settings for the performance tests from https://github.com/alphagov/pay-perftests/blob/master/ci/gatling-simulation-settings.json
5. Runs the 3 test suites we can run without being in the VPC (including the PaymentsSimulation which is the important one)
6. Scales down the test-perf-1 services
7. Stops the test-perf-1 databases
8. unpauses the deploy-to-perf pipeline

It also adds jobs which [allow a developer to scale up or down](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/perf-tests?group=environment-scaling) the test-perf environment and [run the individual perf tests](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/perf-tests?group=individual-perf-tests). Any of these tasks which start/stop the databases will pause the deploy-to-perf pipeline and wait for running jobs to finish to ensure it cannot interfere with running DB migrations.

Finally there is [a pipeline self-update job](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/perf-tests?group=update-perf-tests-pipeline), and a[ job which builds the pay-perftests container ](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/perf-tests?group=perf-tests-build)and uploads it to ECR in test

Should a developer just want to run a full performance test they can simply hit build on the "scale-and-run-all-simulations" job.

I will document all of this in the team manual as well.


ci/tasks/stop-rds-instance.yml
```diff
@@ -7,7 +7,7 @@ image_resource:
 params:
   RDS_INSTANCE_NAME:
   AWS_DEFAULT_REGION: eu-west-1
-  MAX_ATTEMPTS: 100
+  MAX_ATTEMPTS: 140 # 40 minutes
```
jfharden, line 10: Over the weekend there was a failure where it hadn't stopped in the 25 minutes, so give it a bit longer by default

ci/pipelines/perf-tests.yml
```diff
@@ -68,30 +68,485 @@ resources:
...
+      - scale-down-databases
+      - scale-down-services
+      - scale-down-all
   - name: perf-tests-build
     jobs:
       - build-and-push-perf-tests
   - name: update-perf-tests-pipeline
     jobs:
       - update-perf-tests-pipeline
+  - name: all-jobs
```
oswaldquek, line 125: What's the purpose of this "all-jobs"?

oswaldquek (approved):

ci/pipelines/perf-tests.yml
jfharden: Just to see all the things in the serial group together. To be honest I didn't mean to leave it in here :)

jfharden: I'll remove it in one of the follow on

PR #709, opened 2022-06-01
Pp 8337 scale perf tests based on config file

jfharden:
Load the config file created in https://github.com/alphagov/pay-perftests/pull/115 to set scale up and down volumes.

### DIFF
--- a/ci/pipelines/perf-tests.yml
+++ b/ci/pipelines/perf-tests.yml
@@ -572,6 +572,17 @@ jobs:
       - <<: *scale-up-services
       - <<: *prepare-codebuild
       - try:
+          # The payment simulation should be the first simulation run to give time for the events queues to be consumed
+          # completely while the other simulations run. This enables the automatic expiry and expunging processes to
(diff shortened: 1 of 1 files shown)


ci/pipelines/perf-tests.yml
```diff
@@ -145,13 +145,13 @@ definitions:
         params:
           command: scale
           app_name: pay-stubs
-          instances: 1
+          instances: ((.:service-volumes.scale_down_to.stubs))
```
oswaldquek, line 148: Where is this `service-volumes` variable being loaded from?
Also pretty neat that we can define variables in yml files and load them like this.

oswaldquek (approved):

ci/pipelines/perf-tests.yml
jfharden: Yeah, it's really cool :)

I do wish they had a set-var task as well as load, sadly not, but this is pretty neat still!

PR #719, opened 2022-06-17
PP-9741: Display frontend and products-ui and require passing of their smoke tests before push to staging

jfharden:
When the nginx proxy is released currently it gets promoted from test to staging without ever being smoke tested.

Change the nginx-proxy group in deploy-to-test to show the deploys of frontend and products-ui as well as the smoke test jobs for both of these.

Require the passing of both frontend and products-ui smoke tests before promoting the nginx proxy to the staging env.

I did experiment with adding all of the things triggered by an nginx release in test to the view, but that is every service and the view is terrible, this view at least gives us a clear visual of the progress of the container image from release to staging.

You can see the new group at https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test?group=nginx-proxy

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -902,7 +902,10 @@ groups:
   - name: nginx-proxy
     jobs:
       - build-and-push-nginx-proxy-to-test-ecr
-      - deploy-toolbox
+      - deploy-frontend
+      - deploy-products-ui
+      - smoke-test-frontend
+      - smoke-test-products-ui
       - push-nginx-proxy-to-staging-ecr
   - name: webhooks-egress
     jobs:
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -1718,6 +1721,7 @@ jobs:
         trigger: true
       - get: nginx-proxy-ecr-registry-test
         trigger: true
+        passed: [build-and-push-nginx-proxy-to-test-ecr]
```
oswaldquek, line 1724: I was wondering how this would affect the [frontend group](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test?group=frontend) but it appears to be [ok](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-frontend/builds/299).

PR #729, opened 2022-07-12
BAU: Exclude version check from test-perf-1 DB migrations

katstevens:
We ran into problems with this as the image tagged 2856-perf-db had not been deployed as an app release (which was still on 2855-perf).

We don't carry out version checks for test-perf-1 deployments elsewhere, as it's low risk given the perf environment should be up to date with production deployments.

The alternative would be some more complex changes to our application deployment script, or an additional app deployment job just for the *-perf-db tagged images (which would probably cause different problems for this check anyway).

### DIFF
--- a/ci/scripts/run-ecs-db-migration.js
+++ b/ci/scripts/run-ecs-db-migration.js
@@ -93,7 +93,7 @@ const run = async function run () {
     console.log(`Current task definition is using release: ${currentAppRelease}`)
     const jobAppRelease = APPLICATION_IMAGE_TAG.split('-')[0]
 
-    if (currentAppRelease !== jobAppRelease) {
+    if (currentAppRelease !== jobAppRelease && CLUSTER_NAME != "test-perf-1-fargate") {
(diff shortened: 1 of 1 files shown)


ci/scripts/run-ecs-db-migration.js
```diff
@@ -93,7 +93,7 @@ const run = async function run () {
     console.log(`Current task definition is using release: ${currentAppRelease}`)
     const jobAppRelease = APPLICATION_IMAGE_TAG.split('-')[0]
 
-    if (currentAppRelease !== jobAppRelease) {
+    if (currentAppRelease !== jobAppRelease && CLUSTER_NAME != "test-perf-1-fargate") {
```
jfharden, line 96: Just a very minor thing but can you make it a !== so it is type checked too. I know it's bonkers but I do not trust JS even a little bit not to do some type shenanigans and make this be wrong :(

jfharden (approved):

PR #735, opened 2022-08-05
PP-9827: Concourse runner with java 17

oswaldquek:
This is identical to the original
[govukpay/concourse-runner](https://github.com/alphagov/pay-ci/tree/master/ci/docker/concourse-runner)
but with Java 17. The requirement for Java 17 comes from the fact that
[webhooks](https://github.com/alphagov/pay-webhooks) uses it.

Note that docker engine 20.10.17 was originally used but this did not work as dockerd seems to be missing.

This produces a running webhooks integration test: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/webhooks-integration-test/builds/18.

There will be another PR to add this to concourse runner build.


katstevens (approved):
It'd be nice if there was a way to avoid duplication between the two images, but given we rarely update these it's probably ok for the moment.

PR #736, opened 2022-08-05
PP-9292 Run connector provider tests for adminusers PRs

stephencdaly:
- Add adminusers-consumer-pact-test job which runs the consumer contract tests to generate the pact files, then publishes them.
- As part of the adminusers-as-consumer-pact-test, run the connector provider contract tests.
- Run both adminusers-consumer-pact-test and adminusers-integration-test before running adminusers-as-consumer-pact-test, as the  adminusers-consumer-pact-test generates and publishes the pacts with  connector as the producer and adminusers-integration-test generates and  publishes the pacts with ledger as the producer. This can be changed in future so that adminusers-consumer-pact-test will generate both.

Test builds:
https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/adminusers-consumer-pact-test/builds/14
https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/adminusers-as-consumer-pact-test/builds/187

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -101,6 +101,10 @@ definitions:
         path: src
         status: failure
         context: integration tests
+  - &run-java-consumer-pact-tests
+    task: run-consumer-pact-tests
+    privileged: true
+    file: ci/ci/tasks/java-consumer-pact-tests.yml
   - &send-pr-build-time
     task: send-pr-build-time-to-hosted-graphite
     params:
@@ -242,6 +246,7 @@ groups:
       - adminusers-unit-test
       - adminusers-integration-test
       - adminusers-e2e
+      - adminusers-consumer-pact-test
       - adminusers-as-provider-pact-test
       - adminusers-as-consumer-pact-test
       - record-adminusers-build-time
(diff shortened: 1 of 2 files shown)


ci/pipelines/pr.yml
```diff
@@ -1043,14 +1044,67 @@ jobs:
...
+              commit="$(cat .git/resource/head_sha)"
+              pr="$(cat .git/resource/pr)"
+              mvn clean package \
+                        -DPACT_BROKER_URL=https://pact-broker-test.cloudapps.digital \
+                        -DPACT_BROKER_USERNAME="" \
+                        -DPACT_BROKER_PASSWORD="" \
+                        -DPACT_CONSUMER_VERSION="${commit}" \
+                        -DPACT_CONSUMER_TAG="${pr}" \
+                        -Dpact.provider.version="${commit}" \
+                        -DrunConsumerContractTests=true
```
oswaldquek, line 1093: I'm puzzled as to why this works as there is no runConsumerContractTests profile in adminusers's pom.xml

stephencdaly: Ah I've added it in the adminusers PR - I guess probably the adminusers PR needs to come first to add the tests to be run before concourse can run them!

oswaldquek: Also maybe not needed for this PR but if you have time in the future it'd be good to extract this code block as it appears to look like https://github.com/alphagov/pay-ci/blob/master/ci/pipelines/pr.yml#L1417-L1445

stephencdaly: Ah yeh good point, definitely worth doing

oswaldquek (approved):
I feel like "adminusers-consumer-pact-test" and "adminusers-as-consumer-pact-test" are potentially confusing but they are technically the right names!

stephencdaly:
> I feel like "adminusers-consumer-pact-test" and "adminusers-as-consumer-pact-test" are potentially confusing but they are technically the right names!

Yeh they probably are! I copied what was there for ledger, but maybe we could have a think.

PR #740, opened 2022-08-10
PP-9829 smoke test and pact tag stages

nlsteers:
### WHAT

Add smoke testing and pact tagging stages to the `webhooks` test and staging deployment pipelines

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -901,6 +901,8 @@ groups:
     jobs:
       - build-webhooks
       - deploy-webhooks
+      - smoke-test-webhooks
+      - webhooks-pact-tag
       - webhooks-db-migration
       - push-webhooks-to-staging-ecr
   - name: alpine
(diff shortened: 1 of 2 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -5301,6 +5371,38 @@ jobs:
           ENVIRONMENT: test-12
     <<: *put_success_slack_notification      
     <<: *put_failure_slack_notification
+
+  - name: smoke-test-webhooks-egress
```
jfharden, line 5375: Need to add this as a passed param for push-webhooks-egress-to-staging-ecr so the push requires the smoke test to pass

ci/pipelines/deploy-to-test.yml
```diff
@@ -5077,6 +5080,73 @@ jobs:
...
+      - load_var: role
+        file: assume-role/assume-role.json
+        format: json
+      - in_parallel:
+          <<: *smoke-test-run-all-on-test
+    <<: *put_success_slack_notification
+    <<: *put_failure_slack_notification
+
+
+  - name: webhooks-pact-tag
```
jfharden, line 5116: Similarly to the egress this needs to be listed as passed in the push to staging ecr

ci/pipelines/deploy-to-test.yml
nlsteers: Yeah good eyes, I noticed this after applying it.

jfharden (approved):
Crackin' 👍

PR #745, opened 2022-08-19
Pp 9940 build carbon relay ng

jfharden:
This PR adds a build pipeline for the base carbon-relay-ng image.

Efforts are taken to make it as easy as possible to stay up to date with upstream. We are parsing the golang version used to build from the grafana/carbon-relay-ng circleci config.

Despite having already been tested by github this pipeline also runs the tests against the final built container. Considering we are pushing this to a public repository it seems prudent to take the extra care given the container build runs apk upgrade.

A successful run of this pipeline can be seen on concourse https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/carbon-relay-ng/jobs/build-then-test-and-push/builds/4

See the successfully built images in dockerhub: https://hub.docker.com/repository/docker/govukpay/carbon-relay-ng/general (currently private, but we should make this public once we are happy with the build process)

And in ecr in the `govukpay/carbon-relay-ng` repository in the test account

### DIFF
--- a/ci/pipelines/carbon-relay-ng.yml
+++ b/ci/pipelines/carbon-relay-ng.yml
@@ -1,0 +1,243 @@
+---
+resources:
+  - name: pay-ci
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: master
+      username: alphagov-pay-ci-concourse
+      password: ((github-access-token))
+
+  - name: carbon-relay-ng-pipeline
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: master
+      paths:
+        - ci/pipelines/carbon-relay-ng.yml
+      username: alphagov-pay-ci-concourse
+      password: ((github-access-token))
+
+  - name: carbon-relay-ng-src
+    type: git
+    icon: github
+    source:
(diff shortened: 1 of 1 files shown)


ci/pipelines/carbon-relay-ng.yml
```diff
@@ -0,0 +1,243 @@
...
+                echo
+
+                cd carbon-relay-ng-src/govuk-tests
+                echo "Updating docker-compose.yml to use built image"
+                sed -i -E 's/build: ".*"/image: "carbon-relay-ng:test"/' docker-compose.yml
+                echo
+
+                echo "Running tests"
+                docker-compose up --exit-code-from test-receiver
+                echo "Tests passed"
```
katstevens, line 223: Do we want any Slack notifications if the jobs fail?

jfharden: Oh yeah, I think we definitely do

jfharden: I'll set the usual ones up for pass/fail

jfharden:
Added pass/fail notifications.

Pass notification:
Screenshot 2022-08-22 at 16 48 44

I added an exit 1 to one of the tasks to make it fail. The failure:
Screenshot 2022-08-22 at 16 51 06

katstevens (approved):

PR #746, opened 2022-08-25
PP-9792 concourse jobs for webhooks services in production

nlsteers:
### WHAT

`webhooks` deploy-to-production jobs:

- `webhooks`: deploy, smoke test, migrate and pact tag
- `webhooks-egress`: deploy and smoke test

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -477,6 +477,20 @@ resources:
     type: slack-notification
     source:
       url: https://hooks.slack.com/services/((slack-notification-secret))
+  - name: webhooks-ecr-registry-prod
+    type: registry-image
+    icon: docker
+    source:
+      repository: govukpay/webhooks
+      variant: release
+      <<: *aws_prod_config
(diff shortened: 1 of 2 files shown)


nlsteers:
Tested and working

ci/pipelines/deploy-to-production.yml
```diff
@@ -602,6 +616,16 @@ groups:
...
+  - name: webhooks
+    jobs:
+      - deploy-webhooks-to-prod
+      - smoke-test-webhooks-on-prod
+      - webhooks-pact-tag
+      - webhooks-db-migration-prod
+  - name: webhooks-egress
+    jobs:
+      - deploy-webhooks-egress-to-prod
+      - smoke-test-webhooks-egress-on-prod
```
katstevens, line 628: Small request - can we move these groups up, in between `toolbox` and `alpine`? Then it will match staging (I think the test pipeline has `webhooks` in the right place but not `webhooks-egress`)

nlsteers: I've updated the test and prod pipelines

katstevens (approved):

PR #753, opened 2022-09-13
PP-9392 Changed node version to 16

Rkotyank:


ci/pipelines/deploy-to-test.yml
```diff
@@ -2152,7 +2152,7 @@ jobs:
           image_resource:
             type: registry-image
             source:
-              repository: govukpay/node-runner
+              repository: govukpay/node-runner:node16
```
katstevens, line 2155: I think we might have to specify the tag separately:

```
source:
              repository: govukpay/node-runner
              tag: node16
```

There's an example in one of the other tasks: https://github.com/alphagov/pay-ci/blob/a1faf1da14b76ca385081378b7b92c4d93f96610/ci/tasks/run-codebuild.yml#L7

oswaldquek (approved):

PR #761, opened 2022-09-26
PP 10148 remove old node12 version of node runner

Rkotyank:
removed [the symbolic link](https://github.com/alphagov/pay-ci/blob/master/ci/docker/node-runner/Dockerfile) and created a new one to point to Dockerfile.node16 as a default

node12 Dockerfile is removed from pay-ci

[the pay-ci/docker/node-runner README file](https://github.com/alphagov/pay-ci/blob/master/ci/docker/node-runner/README.md) is updated to remove mention of node12

Removed build job for node12 from node-runner.yml

### DIFF
--- a/ci/docker/node-runner/Dockerfile
+++ b/ci/docker/node-runner/Dockerfile
@@ -1,1 +1,1 @@
-Dockerfile.node12
+Dockerfile.node16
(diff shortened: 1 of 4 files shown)


katstevens:
I think the automated check is failing, as the pipeline is still defining a resource that's no longer being used - if you click `Details`, it should link to the Concourse job with the full error message.

Rkotyank:
Yay, the checks have passes

PR #762, opened 2022-10-05
BAU: Increase deploy timeout to 15 minutes

katstevens:
We've seen a significant number of deployments take between 10-13 minutes to show as completed in ECS.

Until we can determine the cause of the delay in achieving the stable state (see PP-10154), let's increase the timeout to 15 minutes, to avoid noisy error messages appearing in Slack for deployments that have actually been successful.


alexbishop1 (approved):

oswaldquek (approved):
I assume each retry is 5 seconds

katstevens:
Yep, 5000ms each retry.

PR #763, opened 2022-10-06
PP-10135: Send release numbers to Hosted Graphite

katstevens:
Adds jobs in the `deploy-to-test` pipeline to send metrics for each app to Hosted Graphite, using a basic netcat command. 

**Notes**
- Concourse already sends metrics to Hosted Graphite for PR build times ([see example for adminusers](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/record-adminusers-build-time/builds/502)). This is done in a separate Concourse resource that runs a baked-in ruby script. I didn't fancy adapting that, hence using a basic alpine container and using `nc` ([as per the Hosted Graphite docs](https://www.hostedgraphite.com/docs/sendingmetrics/supported-protocols.html#tcp-connection)). I did however keep the `record-*` naming scheme for the job.
- The metric needs to be a numeric value, so we can't send over git commits unfortunately.
- The metrics expire eventually, we might want to regularly trigger these jobs (daily?) to ensure the dashboard is populated
- I tried this with a Node script (with a crazy dream about being able to test the code) but it didn't work 😿 
- It's very verbose but that's unavoidable unless we go down the anchor route (which we've so far avoided in this pipeline)
- I haven't added jobs for things like carbon-relay, stunnel, scheduled tasks or the sidecars, as these are a) not updated very often b) slightly more difficult to usefully visualise in HG

Example run for the webhooks app here: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/record-webhooks-release-number/builds/3

You can see a dashboard of the release numbers for test-12 here: https://www.hostedgraphite.com/07ab03f8/grafana/d/O6R0iQVVk/releases?orgId=2&from=now-1h&to=now

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -1481,7 +1481,7 @@ jobs:
           ENVIRONMENT: "test-12"
           APP_NAME: "toolbox"
           RELEASE_NUMBER: ((.:application_image_tag))
-          HOSTED_GRAPHITE_API_TOKEN: ((pr-ci/HOSTED_GRAPHITE_API_KEY))
+          HOSTED_GRAPHITE_API_TOKEN: ((HOSTED_GRAPHITE_API_KEY))
 
   - name: push-toolbox-to-staging-ecr
     plan:
@@ -1563,7 +1563,7 @@ jobs:
           ENVIRONMENT: "test-12"
           APP_NAME: "egress"
           RELEASE_NUMBER: ((.:application_image_tag))
-          HOSTED_GRAPHITE_API_TOKEN: ((pr-ci/HOSTED_GRAPHITE_API_KEY))
+          HOSTED_GRAPHITE_API_TOKEN: ((HOSTED_GRAPHITE_API_KEY))
 
   - name: smoke-test-egress
     serial_groups: [smoke-test]
@@ -1863,7 +1863,7 @@ jobs:
           ENVIRONMENT: "test-12"
           APP_NAME: "frontend"
           RELEASE_NUMBER: ((.:application_image_tag))
-          HOSTED_GRAPHITE_API_TOKEN: ((pr-ci/HOSTED_GRAPHITE_API_KEY))
+          HOSTED_GRAPHITE_API_TOKEN: ((HOSTED_GRAPHITE_API_KEY))
 
   - name: smoke-test-frontend
     serial_groups: [smoke-test]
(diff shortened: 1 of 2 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -1452,6 +1467,36 @@ jobs:
...
+          platform: linux
+          image_resource:
+            type: registry-image
+            source:
+              repository: alpine
+          params:
+            ENVIRONMENT: "test-12"
+            APP_NAME: "toolbox"
+            RELEASE_NUMBER: ((.:application_image_tag))
+            HOSTED_GRAPHITE_API_TOKEN: ((pr-ci/HOSTED_GRAPHITE_API_KEY))
```
jfharden, line 1490: We shouldn't reuse the token from another pipeline in this one, this being prefixed with pr-ci means the token is supposed to be scoped to the pr-ci pipeline (weirdly it feels like a bug that we can even get it!)

jfharden:
Overall I LOVE it, just a couple of suggestions

ci/pipelines/deploy-to-test.yml
```diff
@@ -1452,6 +1467,36 @@ jobs:
...
+  - name: record-toolbox-release-number
+    serial_groups: [hosted-graphite]
+    plan:
+      - get: toolbox-ecr-registry-test
+        trigger: true
+        passed: [deploy-toolbox]
+      - get: pay-ci
+      - load_var: application_image_tag
+        file: toolbox-ecr-registry-test/tag
+      - task: send-release-info-to-hosted-graphite
```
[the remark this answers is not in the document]
jfharden: Having said that, it being it's own job means we can trigger it to send on a schedule which is probably better overall given metric expiry, so forget about making it a task in a job, :) Just the factoring out into a task file

ci/pipelines/deploy-to-test.yml
```diff
@@ -1452,6 +1467,36 @@ jobs:
     <<: *put_success_slack_notification
     <<: *put_failure_slack_notification
 
+  - name: record-toolbox-release-number
+    serial_groups: [hosted-graphite]
```
jfharden, line 1471: I don't think there's any need for these to be in a serial group, feels like creating a bottleneck without a benefit. HG is expecting to receive vast quantities of metrics so it's not like we'll hit a rate limit etc

ci/pipelines/deploy-to-test.yml
katstevens: Oh blimey I forgot I'd done this - good spot, will fix

ci/pipelines/deploy-to-test.yml
katstevens: Agreed, I was being overcautious when I first added in everything (in case they ALL failed! 🔥 ) Will remove

ci/pipelines/deploy-to-test.yml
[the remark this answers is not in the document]
katstevens: Yeah I had it as a step in the deploy job at first, but I couldn't be bothered to wait for all the other steps when testing! Agreed on the scheduling and putting it into a task file though.

katstevens:
@jfharden I think I've resolved all the suggestions now - going to leave the scheduling until I've done staging and production as well, in case we change our minds about how often to do it 😸

jfharden (approved):
📈 Ace!

PR #768, opened 2022-10-17
PP-10050 Added trigger:true to pay-infra-src resource in pay-ci repo

Rkotyank:
Added trigger:true to pay-infra-src resource in pay-ci repo


katstevens (approved):
Looks good to me! 👍

PR #769, opened 2022-10-17
PP-10051 Add base 32 OTP key for e2e tests

stephencdaly:
We are in the process of removing support for non-base 32 encoded OTP keys. Add a new environment variable for a base 32 encoded OTP key for use by the end-to-end tests. The old SELFSERVICE_OTP_KEY environment variable will be removed when end-to-end tests are using the new environment variable.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -230,7 +230,7 @@
         "filename": "ci/tasks/endtoend/docker-config/endtoend.env",
         "hashed_secret": "778c0204b40447fe46db8608ae4718dcd497099f",
         "is_verified": false,
-        "line_number": 37,
+        "line_number": 38,
         "is_secret": false
       }
     ],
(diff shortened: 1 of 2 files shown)


sfount (approved):
Change looks good 👍 

We might want to revert where this was added and wont be used although that might end up as part of a larger cleanup

I wonder if it's worth a light touch README on some of these folders and config files to reference what they're actually consumed by and connect some of the dots

Appreciate that that can change over time and it may not be 100% accurate even knowing what they're originally intended to be consumed by might be helpful lol

For example most of the things here are used by pay-endtoend which might not be immediately clear https://github.com/alphagov/pay-endtoend/blob/293b68a639113f0add00ca1906c01b155343c499/src/test/java/uk/gov/pay/acceptance/utils/ConfigUtils.java

PR #773, opened 2022-10-31
PP-10264: Temporarily disable Dockerhub push steps

katstevens:
We've added MFA to the Concourse Dockerhub user so we have to change the auth method. In the meantime, disable these steps so the pipelines don't break.

The commented-out steps will shortly be reinstated, but with use of an auth token.


oswaldquek (approved):
LGTM as we got past the unauthorized error in https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/selfservice-e2e/builds/641

PR #775, opened 2022-11-01
PP-10254 Pin alpine image resources to 3.16

katstevens:
For ease of reviewing I've split the commits into occurrences of alpine in:
- tasks
- pipelines
- zip-build file (originally used in E2E tests, but is no longer used and has been removed)


jfharden (approved):
👍

PR #777, opened 2022-11-02
PP-10116 add GH action for secrets detection

nlsteers:
### WHAT

- added custom GH action to run `detect-secrets` against an established baseline
- fixed some yml formatting
- added `detect-secrets base` image definition
- concourse pipeline to manage build and push of `detect-secrets` image

### DIFF
--- a/actions/detect-secrets/Dockerfile
+++ b/actions/detect-secrets/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:alpine3.16
+
+RUN apk update && apk upgrade
+RUN apk add bash git
+RUN pip install detect-secrets
+
+COPY entrypoint.sh /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/actions/detect-secrets/Dockerfile
+++ b/actions/detect-secrets/Dockerfile
@@ -1,8 +1,4 @@
-FROM python:alpine3.16
-
-RUN apk update && apk upgrade
-RUN apk add bash git
-RUN pip install detect-secrets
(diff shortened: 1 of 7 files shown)


actions/detect-secrets/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:alpine3.16
```
katstevens, line 1: Can we be a bit more specific? eg `python:3.9-alpine3.16` (see [Dockerhub](https://hub.docker.com/layers/library/python/3.9-alpine3.16/images/sha256-55bc2d88cfd5ed07d2e9420b8f83a1adc4caf49e855b5dd6437d679fae56fb3a?context=explore))

ci/docker/detect-secrets/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM python:3.11.0-alpine3.16@sha256:610aac6d972a31b2d4174a8b260e2cdf69de2d73cf03f4131ca24e61157f34d6
+
+RUN apk update && apk upgrade
+RUN apk add bash git
```
jfharden, line 4: That script will work as plain sh, you could just change the #! to `#!/bin/sh` since it's all POSIX compatible and then you won't need to install bash

nlsteers: good to know, I'd like to keep it though even just to help with future debugging / dev efforts if we revisit this action (and it's only a 1MB difference!)

katstevens (approved):

PR #779, opened 2022-11-07
PP-10264: Unpin concourse-runner image version

katstevens:
The digest changed when we moved the image to the new governmentdigitalservice Docker Hub account.

We don't pin to an image digest elsewhere in the Concourse tasks - I think we can remove this, to streamline the process when we update the image in future.


jfharden (approved):

jfharden:
Agree, especially since it's coming from our own repos!

PR #781, opened 2022-11-08
Pp 10218 update sha256sum of fly for runner 17 and fix dependabot

jfharden:
Bump to the latest docker source image, and also update the sha of fly in the node17 version of the container.

I'm not sure why we have 2 different dockerfiles for this, we should use a build arg to pick the java version and build both from the same Dockerfile.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,4 @@
+---
 version: 2
 updates:
 - package-ecosystem: docker


jfharden:
Had to rebase, @nlsteers also fixed the dependabot config and merged before me 👍

katstevens (approved):

PR #807, opened 2023-01-04
PP-10565: Remove dependabot restriction

jfharden:
Now we have successfully updated to docker 20.10.22 using the `20.10.22-dind` image variant we can remove the dependabot restrictions


katstevens (approved):
Hurray 🎉

PR #816, opened 2023-02-06
PP-10413 add products ui demo payment return url

SandorArpa:
- add a new env var for demo payments return url that was changed due to removing gateway cookies from selfservice


stephencdaly (approved):
LGTM!

PR #823, opened 2023-02-15
PP-10667 semver github action

nlsteers:
### WHAT

Added a new action to determine the correct semantic version to tag a new release with given two possible values
- the action will patch the semantic version by 1 if there is an existing release value that is greater than the source version
- the action will use the source value if is higher than the latest release version
- the action will patch the semantic version by 1 if the source and existing release versions are the same

The action runs on Node 16x as that is the most recent runtime provided by GHA at time of writing


nlsteers:
Example workflow run with this action:

https://github.com/nlsteers/pay-product-page/actions/runs/4186749250/jobs/7255671684

actions/next-semver/.nvmrc
```diff
@@ -0,0 +1 @@
+16.17.1
```
oswaldquek, line 1: Why not 18.12.1?

actions/detect-secrets/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
oswaldquek, line 1: Do we need all this licence stuff?

oswaldquek: And do we need this detect-secrets action? Seems tangential to semver.

actions/next-semver/action.yml
```diff
@@ -0,0 +1,16 @@
...
+    description: 'version in the project source (pom.xml, package.json etc)'
+    required: true
+  release_version:
+    description: 'version of latest github release'
+    required: true
+outputs:
+  version:
+    description: 'semantic version value of the next release'
+runs:
+  using: 'node16'
```
oswaldquek, line 15: Why not node 18? It's node 18 [here](https://github.com/alphagov/pay-ci/pull/823/files#diff-69f6e913dda2dfe289d5b99cb0273ae4fb25629be87ec94285347c241ad867f4R17).

actions/next-semver/dist/index.js
```diff
@@ -0,0 +1,2878 @@
+/******/ (() => { // webpackBootstrap
```
oswaldquek, line 1: Do we need the `actions/next-semver/dist` folder to be checked in?

actions/next-semver/dist/licenses.txt
```diff
@@ -0,0 +1,73 @@
+@actions/core
```
oswaldquek, line 1: Do we need the `actions/next-semver/dist folder` to be checked in?

actions/next-semver/index.ts
```diff
@@ -0,0 +1,59 @@
+// noinspection DuplicatedCode
```
oswaldquek, line 1: The developer in me says it's probably a good idea to have tests for the functions in this file.

oswaldquek (changes requested):

actions/next-semver/.nvmrc
nlsteers: 16x is the latest available GHA runtime for node, see https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-javascript-actions

actions/detect-secrets/LICENSE
nlsteers: this is a publicly available function so seemed prudent to include a license file

actions/next-semver/action.yml
nlsteers: those are the type definitions needed for typescript, not runtime-specific. See https://github.com/alphagov/pay-ci/pull/823#discussion_r1108310292

actions/next-semver/dist/index.js
nlsteers: yes, the action is compiled to a single runnable js file for any consuming workflows. We could probably introduce some sort of pre step in the action.yml to compile at run but this would slow the action down considerably

actions/next-semver/dist/licenses.txt
nlsteers: see https://github.com/alphagov/pay-ci/pull/823#discussion_r1108313431

actions/next-semver/index.ts
nlsteers: I'll add some tests

oswaldquek:
Looks great. What's this annotation error at https://github.com/alphagov/pay-ci/actions/runs/4196532620 though? Can it be gotten rid of?

oswaldquek (approved):

PR #828, opened 2023-02-28
PP 10733 add concourse metrics on the deploy to staging pipeline adminusers

Rkotyank:
Added tasks to get metrics in the deploy-to-staging Adminusers in the following jobs:

- deploy-adminusers-to-staging
- smoke-test-adminusers-on-staging
- adminusers-pact-tag
- push-adminusers-to-production-ecr

### DIFF
--- a/ci/pipelines/deploy-to-staging.yml
+++ b/ci/pipelines/deploy-to-staging.yml
@@ -1891,7 +1891,7 @@ jobs:
         passed: [smoke-test-adminusers-on-staging]
         trigger: true
       - load_var: application_image_tag
-        file: adminusers-ecr-registry-staging/tag 
+        file: adminusers-ecr-registry-staging/tag
       - get: pay-ci
       - task: send-job-release-info-to-hosted-graphite
         file: pay-ci/ci/tasks/send-job-release-info-to-hosted-graphite.yml
@@ -1933,6 +1933,16 @@ jobs:
           format: oci
         trigger: true
         passed: [adminusers-pact-tag]
+      - load_var: application_image_tag
+        file: adminusers-ecr-registry-staging/tag 
+      - get: pay-ci
+      - task: send-job-release-info-to-hosted-graphite
+        file: pay-ci/ci/tasks/send-job-release-info-to-hosted-graphite.yml
+        params:
+          BUILD_PIPELINE_NAME: deploy-to-staging
+          BUILD_JOB_NAME: push-adminusers-to-production-ecr
+          RELEASE_NUMBER: ((.:application_image_tag))
+          HOSTED_GRAPHITE_API_TOKEN: ((HOSTED_GRAPHITE_API_KEY
       - put: adminusers-ecr-registry-prod
         params:
           image: adminusers-ecr-registry-staging/image.tar
--- a/ci/pipelines/deploy-to-staging.yml
+++ b/ci/pipelines/deploy-to-staging.yml
@@ -1933,16 +1933,6 @@ jobs:
           format: oci
         trigger: true
         passed: [adminusers-pact-tag]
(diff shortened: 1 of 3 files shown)


jfharden (commented):
The ci/tasks/parse-staging-release-tag.yml and ci/scripts/parse-staging-release-tag.js files don't seem to be used? Could they be removed or maybe is something that uses them missing from the PR?

Rkotyank:
> The ci/tasks/parse-staging-release-tag.yml and ci/scripts/parse-staging-release-tag.js files don't seem to be used? Could they be removed or maybe is something that uses them missing from the PR?

yes, that was my doubt as well. I don't think we are using it anywhere in this PR or will need to use it for the other apps. we can remove them I think.

Rkotyank:
> > The ci/tasks/parse-staging-release-tag.yml and ci/scripts/parse-staging-release-tag.js files don't seem to be used? Could they be removed or maybe is something that uses them missing from the PR?
> 
> yes, that was my doubt as well. I don't think we are using it anywhere in this PR or will need to use it for the other apps. we can remove them I think.

@katstevens what is your suggestion?

katstevens:
So we need those scripts because the `application_image_tag` value that we're currently passing through to the metrics task is a string, not a numerical value. We need to slice off the `-release` before we can send it to Hosted Graphite.

Rkotyank:
> So we need those scripts because the `application_image_tag` value that we're currently passing through to the metrics task is a string, not a numerical value. We need to slice off the `-release` before we can send it to Hosted Graphite.

hmm, that's probably why I'm not seeing the metrics in HG. let me see where I need to add that. Thanks Kat

ci/pipelines/deploy-to-staging.yml
```diff
@@ -1845,8 +1852,19 @@ jobs:
         trigger: true
         passed: [deploy-adminusers-to-staging]
       - get: pay-ci
+      - task: parse-staging-release-tag
+        file: pay-ci/ci/tasks/parse-staging-release-tag.yml
+        input_mapping:
+          git-release: endtoend-git-release
```
katstevens, line 1858: So these two lines are where the problem is. The  `parse-staging-release-tag.yml` task is currently expecting `inputs` of two folders: `pay-ci` and `ecr-repo`. That's not what we're providing it here 😸

Rkotyank: @katstevens I made this change
 ```
task: parse-staging-release-tag
        file: pay-ci/ci/tasks/parse-staging-release-tag.yml
        input_mapping:
          ecr-repo: adminusers-ecr-registry-staging
```

with a new error 

```
selected worker: ip-10-1-10-50
An error occurred: ReferenceError: release is not defined
```

File: ci/tasks/parse-staging-release-tag.yml [hunk 2]
```diff
@@ -0,0 +1,20 @@
+# This task parses the staging release candidate tag from an ecr resource and turns it into a *-staging release tag
+#
+# File written:
+#   parse-staging-release-tag/tag : The staging release tag (e.g. 123-release)
+#
```
katstevens, line 5: This comment isn't quite describing what's happening in the task, and we should probably update it to reflect the different reason for its use. 

The old task we copied this code from ([parse-perf-release-tag.yml](https://github.com/alphagov/pay-ci/blob/master/ci/tasks/parse-perf-release-tag.yml)) was designed to take a tag like `123-release` as an input, and output a different kind of tag: `123-perf`. Here, we're taking a tag `123-release`, and outputting the release number, `123`. Let's change the explanatory sentence above and the example of what's written to the output file.

katstevens:
Overall this is looking good - I have requested a change to a comment on one of the files though.

PR #834, opened 2023-03-09
PP 10768 adding pact broker to docker

Rkotyank:
- Created ECR repo for pact-broker in the deploy account.

- Updated the [existing pact-broker pipeline](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/pact-broker) to build and push a pact-broker image into the ECR repository in the deploy account.


ci/pipelines/pact-broker.yml
```diff
@@ -35,6 +41,40 @@ resources:
...
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: master
+
+  - name: pact-broker-ecr-registry-test
+    type: registry-image
+    icon: docker
+    source:
+      repository: govukpay/pact-broker
+      tag: "2.54.0.0"
```
katstevens, line 65: I think we should probably add a comment here to say that we're hardcoding this tag for now (and give a short summary of the reason why - we don't want to upgrade to a new version of pact broker while we're in the process of moving it). Once we're ready to upgrade we'll need to think about how we'll get the value for this `tag` field (this is a separate story).

Though just typing that, I've realised that Dependabot is going to try and upgrade the version in the `Dockerfile` as soon as we merge this PR... and it'll be tagged wrongly as 2.54.0.0! I'll make a note to snooze that Dependabot PR when it appears.

ci/docker/pact-broker/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM pactfoundation/pact-broker:2.54.0.0
```
katstevens, line 1: Some EOF characters have sneaked into this file and the one above - adding a new line to each file should clean that up.

katstevens (approved):
Looks great 👍

katstevens (approved):
Lovely clean commits on this PR 👍 ❤️

PR #863, opened 2023-04-24
BAU: Switch Concourse PR node tests to use alpine3.17

katstevens:
16-stretch-slim is no longer available. ~We use Alpine base images for the actual Node app images, so one would hope their tests should run OK on `node:16-alpine3.17`...~ We're going to try `bullseye-slim` instead for now, as the Alpine change would require some extra wrangling.


alexbishop1 (approved):
🤷 I mean, it can’t be any more broken than it already is

jfharden (approved):

PR #892, opened 2023-06-05
PP-8143: fully continuously deploy all apps and sidecars

jfharden:
Enable full continuous deployment for all application images and sidecars out to production.

This was agreed in [this ADR](https://github.com/alphagov/pay-architecture/blob/main/adr/021-continuously-deploy-all-applications.md)

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -688,6 +688,7 @@ jobs:
     serial_groups: [deploy-application]
     plan:
       - get: egress-ecr-registry-prod
+        trigger: true
       - get: pay-infra
       - get: pay-ci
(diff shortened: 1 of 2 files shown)


jfharden:
It's true the sidecars do trigger a release of everything at once, but isn't that desirable? Deploys don't take long now (back when it was a problem every deploy took a minimum of about 6 minutes, now it's 1 and a half apart from egress.

jfharden:
Not having it leaves us in a similar position to currently where a seemingly random amount of things will get deployed (it's perfectly faesible we would deploy a new frontend feature and at the same moment also deploy new nginx-forward-proxy, nginx-proxy, and telegraf sidecars, which seems like a bad position to be in?

katstevens (approved):

PR #905, opened 2023-07-07
PP-11168: Add docker scout concourse task

jfharden:
Experiment for using docker scout as part of our DevSecOps strategy


jfharden:
We've decided against doing this for the moment

PR #917, opened 2023-07-26
PP-11224: Push ADOT image to production ECR

oswaldquek:
The pipeline can be seen in [concourse](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/deploy-to-staging/jobs/push-adot-to-production-ecr/builds/1).

Given there is an [ADOT repo in prod ECR](https://github.com/alphagov/pay-infra/blob/master/provisioning/terraform/deployments/production/account/ecr.tf) I have surmised that this pipeline should exist.


ci/pipelines/deploy-to-staging.yml
```diff
@@ -545,6 +559,9 @@ resource_types:
       tag: latest
 
 groups:
+  - name: adot
```
jfharden, line 562: This group needs to be later in the groups. The order of the groups is the order in the UI and for every other pipeline is:

* all node & java apps in alphabetical order; then
* all sidecars in alphabetical order

So it should appear after webhooks-egress  and before alpine

jfharden:
You're right this needs to exist, we also need to do the same thing to push from prod to perf.

I've added that to the PR and it can be seen working in https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/deploy-to-production?group=adot

PR #918, opened 2023-08-02
PP-11292: Update pact-broker urls

jonathan-ryding:
Update pact-broker urls within Concourse so that pact tests, new version tags, pact publishing are done using the migrated pact-broker once its database is updated.

The pactbroker url is changing from https://pay-pact-broker.cloudapps.digital/ & https://pact-broker-test.cloudapps.digital/ [[1](https://github.com/alphagov/pay-ci/blob/711d25cdfee3a47e3e58845badc25f66d26735cd/ci/pact_broker/pay-pact-broker_manifest.yml#L19C1-L21C48)] to https://pact-broker.deploy.payments.service.gov.uk/

### DIFF
--- a/ci/tasks/check-pact-compatibility.yml
+++ b/ci/tasks/check-pact-compatibility.yml
@@ -16,7 +16,7 @@ run:
     - -ec
     - |
       pact-broker can-i-deploy \
-        --broker-base-url https://pact-broker-test.cloudapps.digital \
+        --broker-base-url https://pact-broker.deploy.payments.service.gov.uk/ \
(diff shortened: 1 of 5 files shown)


oswaldquek (approved):
Looks good. This should only be applied during the code freeze window when the DB is switched over to the AWS RDS one. 
Thing to do once the DB has been switched over: re-run a deploy (or merge a dependabot PR) to verify the pact functions (i.e. checking pact compatibility, publishing to pact broker) works.
Come up with a backup plan in case anything goes wrong and you need to revert.

PR #920, opened 2023-08-08
PP-11226: Update deploy-to-test to deploy adot

jfharden:
Quite a lot of stuff had to change to make this possible so I've only updated deploy-to-test.
The adot sidecar is still disabled so this PR doesn't cause the apps to get adot deployed with them, only make it possible to do so.

I had to update various scripts, and in some cases it was better to copy the script and add the extra ADOT_IMAGE_TAG code. When we are removing telegraf we can remove the script duplication and go back to just having deploy-app.yml and deploy-frontend.yml.

Future updates will be smaller since nearly all the task files will have been updated in this PR.

I applied this pipeline and you can see an example of a working job here: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/deploy-products-ui/builds/624 since trigger:true was on for the adot containers it actually ran all the java and node service deploys so they have all been tested.


ci/tasks/deploy-frontend-with-adot.yml
```diff
@@ -0,0 +1,36 @@
+---
```
jfharden, line 1: Note this file is a copy of deploy-app.yml just with the addition of the adot variables

ci/tasks/create-notification-snippets.yml
```diff
@@ -41,6 +42,13 @@ run:
       echo "- <https://github.com/alphagov/pay-nginx-proxy/releases/tag/alpha_release-${NGINX_RELEASE_NUMBER}|nginx-proxy v${NGINX_IMAGE_TAG}>" | \
       tee -a snippet/failure || true
 
+      # Disable the shellcheck rull telling us this isn't an if-then-else. We know, and it's the desirable
```
oswaldquek, line 45: "rule"

oswaldquek (approved):

ci/tasks/create-notification-snippets.yml
jfharden: Ha good spot, this same comment is copied loads of times in this file 😆 

I'll update the misspelling in the follow on PR for staging

PR #930, opened 2023-08-29
BAU: Remove pact test serial group

jfharden:
Remove the pact-test serial group.

There's no reason I know of to require the pact tests to run serially across different projects.

Having them in a serial group is slowing PR tests and causing serial conflicts with concourse getting confused and occasionally never running any jobs.

Note: I have already applied this to test the theory that it was the serial group stopping the jobs from running, and to unblock developers

### DIFF
--- a/ci/pipelines/pr.yml
+++ b/ci/pipelines/pr.yml
@@ -512,7 +512,6 @@ jobs:
       put: webhooks-pull-request
 
   - name: webhooks-as-consumer-pact-test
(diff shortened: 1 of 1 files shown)


oswaldquek (approved):

jfharden:
@katstevens oh that's interesting. It's possible if loads of prs all open at 3am this could happen. We can always try and stagger the dependabot times across repos

PR #933, opened 2023-08-30
PP-11227-add-adot-sidecar-to-cd

rfisher-gds:
## What?
We don't currently test the full metrics path when we make a deployment. This PR exercises the ADOT sidecar and asserts that metrics from a test Toolbox deployment make it through, correctly tagged, to our hosted Prometheus.

## How?
Adds a `run-adot-integration-test` job to the `deploy-to-test` pipeline. This depends on a successful build of the ADOT sidecar, which it attaches to a `toolbox` in the test account. A script is run via `node-runner`, which connects to Amazon Managed Prometheus and performs and instant query, asserting that as of "right now" Toolbox in the test environment is sending metrics. Note: **this does not currently assert a particular version of Toolbox**, which may be incorrect. If the test finds metrics, it passes and the pipeline continues. If not, the test fails and halts the pipeline.

This change has required other changes [to grant Concourse privilege to access Prometheus](https://github.com/alphagov/pay-concourse-deployment/pull/100), and to [build the AWS4 module into `node-runner`](https://github.com/alphagov/pay-ci/pull/936).

Example of successful test job: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/run-adot-integration-test/builds/26

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -1527,7 +1527,7 @@ jobs:
       - get: nginx-proxy-ecr-registry-test
         trigger: true
       - get: adot-candidate
-        passed: [deploy-toolbox]
+        passed: [build-and-push-adot-candidate]
         trigger: true
       - get: pay-infra
       - get: pay-ci
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -1572,31 +1572,6 @@ jobs:
         params:
           APP_NAME: toolbox
           <<: *wait_for_deploy_params
-      - task: run-toolbox-metric-tests
-        config:
-          platform: linux
-          image_resource:
-            type: registry-image
-            source:
-              repository: governmentdigitalservice/pay-node-runner
-              tag: node16
-          inputs:
-            - name: pay-ci
-          params:
-            TEST_METRIC_IMAGE_TAG: candidate
-            TEST_METRIC_ECS_SERVICE: toolbox
-            # TODO: find out how to get endpoint and creds
-            # AWS_ACCESS_KEY_ID: ((.:role.AWS_ACCESS_KEY_ID))
-            # AWS_SECRET_ACCESS_KEY: ((.:role.AWS_SECRET_ACCESS_KEY))
-            # AWS_SESSION_TOKEN: ((.:role.AWS_SESSION_TOKEN))
-            # AWS_REGION: "eu-west-1"
-            # CLUSTER_NAME: "test-12-fargate"
-            # APP_NAME: "adminusers"
-            # APPLICATION_IMAGE_TAG: ((.:application_image_tag))
-          run:
-            path: node
-            args:
-              - pay-ci/ci/scripts/test-metrics.js
     <<: *put_success_slack_notification
     <<: *put_failure_slack_notification
 
@@ -6620,6 +6595,24 @@ jobs:
               image: adot-candidate/image.tar
           get_params:
             skip_download: true
+      - task: run-toolbox-metric-tests
+        config:
+          platform: linux
+          image_resource:
+            type: registry-image
+            source:
+              repository: governmentdigitalservice/pay-node-runner
+              tag: node16
+          inputs:
+            - name: pay-ci
+          params:
+            TEST_METRIC_IMAGE_TAG: candidate
+            TEST_METRIC_ECS_SERVICE: toolbox
+            <<: *aws_test_config
+          run:
+            path: node
+            args:
+              - pay-ci/ci/scripts/test-metrics.js
     on_failure:
       put: slack-notification
       attempts: 10
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -6576,25 +6576,6 @@ jobs:
             ecr-repo: adot-candidate
         - load_var: candidate_image_tag
           file: adot-candidate/tag
-      # We'll add an actual integration test later!
-      - in_parallel:
-        - do:
-          - put: adot-ecr-registry-test
-            params:
-              image: adot-candidate/image.tar
-              additional_tags: parse-candidate-tag/release-tag
-            get_params:
-              skip_download: true
-          - put: adot-latest
-            params:
-              image: adot-candidate/image.tar
-            get_params:
-              skip_download: true
-        - put: adot-dockerhub
-          params:
-              image: adot-candidate/image.tar
-          get_params:
-            skip_download: true
       - task: run-toolbox-metric-tests
         config:
           platform: linux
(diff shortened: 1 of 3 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -1526,7 +1526,8 @@ jobs:
         passed: [build-and-push-telegraf-to-test-ecr]
       - get: nginx-proxy-ecr-registry-test
         trigger: true
-      - get: adot-ecr-registry-test
+      - get: adot-candidate
+        passed: [deploy-toolbox]
```
jfharden, line 1530: This job is the deploy-toolbox job, you want to reference the previous job in the chain that adot-candidate would have gone through here, which would be `build-and-push-adot-candidate`

ci/pipelines/deploy-to-test.yml
```diff
@@ -1571,6 +1572,31 @@ jobs:
         params:
           APP_NAME: toolbox
           <<: *wait_for_deploy_params
+      - task: run-toolbox-metric-tests
```
jfharden, line 1575: This task wants to go into the run-integration-tests job rather than the deploy-toolbox job. On that job you can specify adot-candidate requiring deploy-toolbox to have passed

ci/pipelines/deploy-to-test.yml
```diff
@@ -1571,6 +1572,31 @@ jobs:
...
+            type: registry-image
+            source:
+              repository: governmentdigitalservice/pay-node-runner
+              tag: node16
+          inputs:
+            - name: pay-ci
+          params:
+            TEST_METRIC_IMAGE_TAG: candidate
+            TEST_METRIC_ECS_SERVICE: toolbox
+            # TODO: find out how to get endpoint and creds
```
jfharden, line 1588: There's a rather helpful yaml anchor which will fill all this in for you `&aws_test_config` which has all these in:
```
   1 aws_test_config: &aws_test_config
   2   aws_access_key_id: ((readonly_access_key_id))
   3   aws_secret_access_key: ((readonly_secret_access_key))
   4   aws_session_token: ((readonly_session_token))
   5   # Need to create the role for the concourse user to assume in the test account
   6   aws_role_arn: arn:aws:iam::((pay_aws_test_account_id)):role/concourse
   7   # Hardcode the test account registry ID for now. Needs to be a string, not a number
   8   aws_ecr_registry_id: "((pay_aws_test_account_id))"
   9   aws_region: eu-west-1
   ```
   in

jfharden: Endpoint is an interesting one. I think it's probably ok to hard code it which we have done in some other places. The workgroup url can't change or we're right up the swanny :D

ci/pipelines/deploy-to-test.yml
rfisher-gds: Is that right? I only see that job in the PR pipeline.

ci/pipelines/deploy-to-test.yml
rfisher-gds: That's what I thought. I found the endpoint in the Grafana datasource after I put in the "find out how to get endpoint" comment. I thought we'd need to assume a role rather than getting access keys.

ci/pipelines/deploy-to-test.yml
rfisher-gds: Or do you mean add a new `run-integration-tests` job to `deploy-to-test`?

jfharden: Sorry I mean [run-adot-integration-test](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/run-adot-integration-test/builds/6)

ci/pipelines/deploy-to-test.yml
jfharden: The role the concourse worker has doesn't have permission to query prometheus, but we also don't have a role for it to assume to query the metrics 🤔 
I think doing either would be ok, and it's much easier to give concourse the ability to query the prometheus metrics than making a new role just for it.

The concourse worker role is created by this module usage: https://github.com/alphagov/pay-concourse-deployment/blob/main/terraform/deployments/pay-cd/team-pay-dev.tf#L23, you can get the role name for doing a role policy attachment by doing `module.concourse_keys.concourse_worker_iam_role_names["pay-dev"]`. There's an AWS pre-baked policy to query prometheus we can use which we [already attach to grafana here](https://github.com/alphagov/pay-concourse/blob/main/terraform/modules/concourse-monitoring/iam.tf#L326-L333)

rfisher-gds: API calls must be signed, as-per https://docs.aws.amazon.com/prometheus/latest/userguide/AMP-onboard-query-APIs.html. There's no aws-sdk helper for it, you're (to some degree) on your own.

I've added signing to the script, and its `aws-crt` dependency to the node-runner container in #936. Looking at similar jobs, I think the creds required for signing will be passed through to the container.

jfharden:
Looks great, just one request. Could you run this looking for something we know wont be there (like ecsServiceName: foo) and check it errors. And also for the service 'express-service' which [had metrics until about 5 minutes ago](https://grafana.monitoring.pay-cd.deploy.payments.service.gov.uk/explore?orgId=1&left=%7B%22datasource%22:%22af8fd347-d3c4-49d5-aa17-500f027906d1%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7BecsServiceName%3D%5C%22express-service%5C%22%7D%22,%22range%22:true,%22instant%22:false,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22af8fd347-d3c4-49d5-aa17-500f027906d1%22%7D,%22editorMode%22:%22builder%22,%22exemplar%22:false%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D) (hopefully by the time you read this that will be true, I just destroyed it)

rfisher-gds:
> Looks great, just one request. Could you run this looking for something we know wont be there (like ecsServiceName: foo) and check it errors. And also for the service 'express-service' which [had metrics until about 5 minutes ago](https://grafana.monitoring.pay-cd.deploy.payments.service.gov.uk/explore?orgId=1&left=%7B%22datasource%22:%22af8fd347-d3c4-49d5-aa17-500f027906d1%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7BecsServiceName%3D%5C%22express-service%5C%22%7D%22,%22range%22:true,%22instant%22:false,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22af8fd347-d3c4-49d5-aa17-500f027906d1%22%7D,%22editorMode%22:%22builder%22,%22exemplar%22:false%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D) (hopefully by the time you read this that will be true, I just destroyed it)

Yes, it fails correctly. See https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/run-adot-integration-test/builds/27

jfharden (approved):

PR #944, opened 2023-09-20
PP-11411-upgrade-concourse-script-containers-to-node-18

rfisher-gds:
## What
Node 16 is out of support. Move internal tooling to Node 18.

## How
Move Github actions from 16 to 18; build new Node 18 version of node-runner and have Concourse use it; use Node 18 for Concourse PR jobs.

Use Node18 runtime for logging Lambdas. They all pass their unit tests on Node 18.

I have run what scripts I could locally with Node 16 and 18 containers, and observed no difference.

I'VE LOOKED HARD BUT I MAY HAVE MISSED SOMETHING!

### DIFF
--- a/ci/docker/node-runner/Dockerfile.node18
+++ b/ci/docker/node-runner/Dockerfile.node18
@@ -1,4 +1,4 @@
-FROM node:18.17.1-alpine3.18@sha256:3482a20c97e401b56ac50ba8920cc7b5b2022bfc6aa7d4e4c231755770cf892f
+FROM node:18.18.0-alpine3.18@sha256:619ce27eb37c7c0476bd518085bf1ba892e2148fc1ab5dbaff2f20c56e50444d
 
 # As of node 15 the docker container fails to npm install without either a WORKDIR or -g
 WORKDIR /node-runner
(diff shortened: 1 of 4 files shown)


ci/docker/node-runner/Dockerfile.node18
```diff
@@ -0,0 +1,10 @@
+FROM node:18.18.0-alpine3.18@sha256:619ce27eb37c7c0476bd518085bf1ba892e2148fc1ab5dbaff2f20c56e50444d
```
jfharden, line 1: This is an image sha for an arm64 image, but these images need to run on concourse which is amd64 and doesn't have any cross-platform capability.

You can get an amd64 SHA by [searching the node image tags](https://hub.docker.com/_/node/tags?page=1&name=18.18.0-alpine) then clicking through to the `linux/amd64` and getting the full sha at the top of that page

jfharden (commented):
Just need to update the SHA to be an amd64 image

ci/docker/node-runner/Dockerfile.node18
rfisher-gds: Agh! Good spot.

jfharden (approved):

PR #950, opened 2023-10-12
PP-11459-create-stubs-ecs-pipeline

rfisher-gds:
## What

Add a pipeline to build and deploy the Stubs application in ECS.

A couple of things about this pipeline are "old". First, though we have other tickets to move tests into Github Actions, building from the Stubs `Dockerfile` runs tests in-place so, in this case, in Concourse. Second, the tag of the image is always `latest-master`. These don't seem ideal, but I chose to leave them as-is. I have, however, added code to put in build-date and Git SHA labels, in line with other repos.

[The pipeline currently in Concourse](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/stubs) has been deployed from my laptop. The deploy phase is not expected to work, as it currently has nothing to deploy to. The ECS service is defined in https://github.com/alphagov/pay-infra/pull/4547, which is not yet applied.

The Slack notifications are commented out, so as not to add noise whilst things are being debugged. I will uncomment them in a separate PR once we're up and running.

### DIFF
--- a/ci/pipelines/stubs.yml
+++ b/ci/pipelines/stubs.yml
@@ -1,0 +1,200 @@
+---
+resource_types:
+  # - name: cf-cli
+  #   type: docker-image
+  #   source:
+  #     repository: nulldriver/cf-cli-resource
+
+  - name: slack-notification
+    type: docker-image
+    source:
+      repository: cfcommunity/slack-notification-resource
+      tag: latest
+
+resources:
+  - name: stubs-pipeline
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: PP-11459-create-stubs-ecs-pipeline
+      paths:
+        - ci/pipelines/stubs.yml
+
+  - name: stubs-src
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-stubs
+      branch: master
+      paths:
+        - ./*
+
+  - name: pay-ci
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: PP-11459-create-stubs-ecs-pipeline
+
+  - name: pay-infra
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-infra
+      branch: master
+      username: alphagov-pay-ci-concourse
+      password: ((github-access-token))
+
+  - name: stubs-ecr-registry-deploy
+    type: registry-image
+    icon: docker
+    source:
+      repository: govukpay/stubs
+      aws_access_key_id: ((readonly_access_key_id))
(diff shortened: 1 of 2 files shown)


ci/pipelines/stubs.yml
```diff
@@ -0,0 +1,211 @@
...
+  - name: stubs-src
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-stubs
+      branch: master
+      username: alphagov-pay-ci-concourse
+      password: ((github-access-token))
+      paths:
+        - ./*
```
oswaldquek, line 33: Not sure about this block. Maybe one for @jfharden to verify. I think we should be watching on a release tag, similar to [other microservices](https://github.com/alphagov/pay-ci/blob/master/ci/pipelines/deploy-to-test.yml#L160). We're using [release tags on a merge to master for pay-stubs too](https://github.com/alphagov/pay-stubs/actions/runs/6563571278/job/17828136214).

ci/pipelines/stubs.yml
```diff
@@ -0,0 +1,211 @@
...
+
+  - name: pay-ci
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-ci
+      branch: PP-11459-create-stubs-ecs-pipeline
```
oswaldquek, line 40: This will need to be "master"

ci/pipelines/stubs.yml
```diff
@@ -0,0 +1,211 @@
...
+        file: tags/release-sha
+      - load_var: date
+        file: tags/date
+      - task: generate-docker-creds-config
+        file: pay-ci/ci/tasks/generate-docker-config-file.yml
+        params:
+          USERNAME: ((docker-username))
+          PASSWORD: ((docker-access-token))
+          EMAIL: ((docker-email))
+      # The Stubs image in Docker Hub is always tagged 'latest-master'
```
oswaldquek, line 149: I don't think we need this comment; or it could be placed on current line 166.

oswaldquek (changes requested):

ci/pipelines/stubs.yml
jfharden: Yeah rather than any path on master if you could release on the alpha-release_XXX tag instead (this is only applied after all the in-repo tests take place post merge)

jfharden: The only thing of extra note to this:

In the task which builds and pushes to deploy we need to [parse the release tag](https://github.com/alphagov/pay-ci/blob/master/ci/pipelines/deploy-to-test.yml#L1515), and push it to ECR with the XXX-release tag instead of alpha_release-XXX

oswaldquek (approved):
LGTM!

oswaldquek (approved):

PR #954, opened 2023-10-19
PP-11675: Upgrade node-runner, remove Node16

katstevens:
Bring in Node18 security updates.

I've removed the separate Node16/Node18 Dockerfiles and symlink in favour of one Dockerfile - it is less confusing for Dependabot (and us) to understand/update. 

For reference, [this system was introduced in 2022](https://github.com/alphagov/pay-ci/pull/648) when we were upgrading from Node12 to Node16, and needed both versions to ensure backward compatibility.


rfisher-gds (approved):
LGTM

PR #960, opened 2023-11-02
PP-11703: Start perf DBs for weekly maintenance window

jfharden:
Start perf databases during the weekly maintenance window to ensure AWS can apply minor updates and OS security patches etc.

You can see this applied https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/perf-tests

### DIFF
--- a/ci/pipelines/perf-tests.yml
+++ b/ci/pipelines/perf-tests.yml
@@ -95,7 +95,6 @@ resources:
       start: "03:15"
       stop: "03:25"
       days: [Sunday]
-      initial_version: true
 
   - name: every-sunday-at-0450
     type: time
@@ -105,7 +104,6 @@ resources:
       start: "04:50"
       stop: "05:00"
       days: [Sunday]
-      initial_version: true
 
 resource_types:
   - name: slack-notification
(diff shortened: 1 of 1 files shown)


ci/pipelines/perf-tests.yml
```diff
@@ -87,6 +87,24 @@ resources:
       location: Europe/London
     type: time
 
+  - name: every-sunday-at-0315
```
rfisher-gds, line 90: I think it would be better if the name told you what it does rather than when it does it.

rfisher-gds (changes requested):

ci/pipelines/perf-tests.yml
jfharden: What it does though is dependent on what it's connected to. You can connect as many things as you want to this. If you see it in the context of the pipeline 
Screenshot 2023-11-02 at 16 41 29
It's a lot more useful to have it say when, otherwise you can't tell it's a timer.

We have this same nomenclature in various places:

Screenshot 2023-11-02 at 16 42 10
Screenshot 2023-11-02 at 16 42 23

rfisher-gds (approved):

PR #961, opened 2023-11-06
PP-11873-scale-stubs-when-perf-test-run

rfisher-gds:
Scale Stubs in ECS rather than in PAAS when we run the perf-test pipeline.

You can see successful scale-up and scale-downs:
* https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/perf-tests/jobs/scale-up-all/builds/61.15
* https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/perf-tests/jobs/scale-down-all/builds/27.2

This change will not work without the IAM changes in https://github.com/alphagov/pay-infra/pull/4617

### DIFF
--- a/ci/pipelines/perf-tests.yml
+++ b/ci/pipelines/perf-tests.yml
@@ -147,15 +147,24 @@ groups:
 definitions:
   - &scale-down-services
     in_parallel:
-      - task: scale-down-stubs
-        file: pay-ci/ci/tasks/scale-fargate-service.yml
-        params:
-          SERVICE_NAME: stubs
-          SCALE_DIRECTION: in
-          DESIRED_HEALTHY_INSTANCES: ((.:service-volumes.scale_down_to.stubs))
-          AWS_ACCESS_KEY_ID: ((.:role.AWS_ACCESS_KEY_ID))
-          AWS_SECRET_ACCESS_KEY: ((.:role.AWS_SECRET_ACCESS_KEY))
-          AWS_SESSION_TOKEN: ((.:role.AWS_SESSION_TOKEN))
+      - do: # Stubs is in the deploy account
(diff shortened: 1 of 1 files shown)


kbottla (approved):

ci/pipelines/perf-tests.yml
```diff
@@ -158,12 +147,15 @@ groups:
...
+      - task: scale-down-stubs
+        file: pay-ci/ci/tasks/scale-fargate-service.yml
         params:
-          command: scale
-          app_name: pay-stubs
-          instances: ((.:service-volumes.scale_down_to.stubs))
+          SERVICE_NAME: stubs
+          SCALE_DIRECTION: in
+          DESIRED_HEALTHY_INSTANCES: ((.:service-volumes.scale_down_to.stubs))
+          AWS_ACCESS_KEY_ID: ((.:role.AWS_ACCESS_KEY_ID))
```
jfharden, line 156: These wont work because the roles assumed by these creds are in other AWS accounts (the roles assumed are in test, staging, or production), but stubs is in the deploy account

I'd be interested to see if this works without providing the AWS creds (meaning concourse already has the permission it needs). Otherwise you might need to assume a separate role for just this service

rfisher-gds: Are you sure? All the other services are the same.

jfharden: Yeah, all the other services are in the test account, but this ECS service is in the deploy account. Feel free to apply it and test it out though :)

rfisher-gds: Ah, yes!

jfharden (approved):

PR #964, opened 2023-11-09
PP-11785-deploy-stubs-to-perf

rfisher-gds:
Deploy Stubs to ECS rather than PAAS in test-perf.

I'm doing this rather blindly. It _looks_ like the right thing to do, but I need a reviewer more aware of the whole test-perf ecology to tell me I'm on the right path.

### DIFF
--- a/ci/pipelines/deploy-to-perf.yml
+++ b/ci/pipelines/deploy-to-perf.yml
@@ -1588,7 +1588,7 @@ jobs:
         username: pay-concourse
         text: ":red_circle: Deployment of notifications to perf failed - <https://pay-cd.deploy.payments.service.gov.uk/teams/$BUILD_TEAM_NAME/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME|Concourse build #$BUILD_NAME>"
 
-    - name: deploy-scheduled-tasks
+  - name: deploy-scheduled-tasks
     plan:
       - in_parallel:
         - get: alpine-ecr-registry-perf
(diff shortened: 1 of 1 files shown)


jfharden:
So I think we already deploy stubs to fargate in the stubs pipeline: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/stubs

So i think you can actually delete all the stubs stuff from the deploy-to-perf pipeline

For scaling it up and down you want to edit the perf-tests pipeline instead and replace the [scale-up on paas](https://github.com/alphagov/pay-ci/blob/master/ci/pipelines/perf-tests.yml#L415) with an equivalent for ECS (see [scale-up-selfservice](https://github.com/alphagov/pay-ci/blob/master/ci/pipelines/perf-tests.yml#L406) for an example, although that's using an AWS Role in the test account which was loaded earlier in the job, but stubs is in deploy so you'll either need to assume another role (or maybe the concourse worker already has permission given it's in deploy))

Scaling it down is [similar](https://github.com/alphagov/pay-ci/blob/master/ci/pipelines/perf-tests.yml#L161) but with different params

rfisher-gds:
> So I think we already deploy stubs to fargate in the stubs pipeline: https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/stubs
> 
> So i think you can actually delete all the stubs stuff from the deploy-to-perf pipeline

Yes. I kept flip-flopping on this, and eventually decided I didn't quite grasp the way it works yet.

> For scaling it up and down you want to edit the perf-tests pipeline instead 

This is done in #961 , which isn't merged yet.

jfharden (approved):

PR #972, opened 2023-11-22
BAU: Fix default ecr repo names for e2e tests

jfharden:
We changed from the public dockerhub org govukpay to governmentdigitalservice a while back, but these defaults haven't been updated making it harder to run these tests locally.

We also prefixed all the repo names with pay- in the new org to distinguish ourselves from other gds teams repos and prevent repo name collision.

### DIFF
--- a/ci/tasks/endtoend/card/docker-compose.yml
+++ b/ci/tasks/endtoend/card/docker-compose.yml
@@ -6,7 +6,7 @@ networks:
 
 services:
   publicapi:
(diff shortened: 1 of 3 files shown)


katstevens:
I think we need to add the docker credentials to the e2e job in the PR pipeline?

jfharden:
Ignore this PR. The real end to end tests run from ECR not dockerhub, and this PR breaks them because the repos are not similarly named. Instead I'm going to try and add some shell scripts to this repo to run them locally instead

PR #973, opened 2023-11-22
PP-11899: update to localstack 3

jfharden:
The aws-sdk for java uses the sqs JSON api which isn't compatible with localstack version 2.

1. Update to localstack 3 to get that compatibility so that end to end tests can work successfully when the aws-sdk is bumped
2. Add the payout reconcile queue URL into the connector env to stop connector endlessly logging an error
3. Update the e2e-helpers pipeline to copy localstack version 3 to ecr
4. Add a script to make it easier to run the e2e tests locally (and to warn devs the actual tests can't work on an ARM based computer)


SandorArpa (approved):
LGTM

PR #987, opened 2023-12-06
BAU: Move daily perf tests to a more carbon-friendly time

jfharden:
1. Move the nightly perf tests to 4am since it's a more carbon emissions friendly time than 11pm.

Change the other schedules in the perf-tests pipeline to run using Europe/London timezone instead of UTC to ensure the trigger times don't overlap and make more sense when we view them in the UI.


jfharden:
We need better data to understand if this is truly effective.

PR #989, opened 2023-12-13
PP-10386-build-cardid-in-codebuild

rfisher-gds:
Build multi-arch cardid Docker images in CodeBuild, and publish them to ECR and Docker Hub.

This PR has [a companion in `pay-infra`](https://github.com/alphagov/pay-infra/pull/4679) which sets up the IAM role Concourse needs to run CodeBuild jobs.

The work here can easily be repurposed to other Pay applications. N

[Here is a completed Concourse job](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/codebuild-cardid/builds/38) and [here are the published images](https://eu-west-1.console.aws.amazon.com/ecr/repositories/private/223851549868/govukpay/cardid?region=eu-west-1).

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -4450,7 +4450,7 @@ jobs:
               - -ec
               - |
                 apk add --quiet --no-progress git
-                sed -i "s/git@github.com:/https:\/\/${GH_ACCESS_TOKEN}@github.com\//" .gitmodules
+                sed -i "s/https:\/\/github.com/https:\/\/${GH_ACCESS_TOKEN}@github.com\//" .gitmodules
                 git submodule init -q data
                 git submodule update data
                 sed -i "s/${GH_ACCESS_TOKEN}/token_redacted/" .gitmodules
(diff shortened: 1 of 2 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -4415,6 +4416,108 @@ jobs:
...
+            type: registry-image
+            source:
+              repository: alpine
+              tag: 3.16
+          inputs:
+            - name: cardid-git-release
+          params:
+            GH_ACCESS_TOKEN: ((github-access-token))
+          outputs:
+            - name: cardid-git-release-with-submodules
```
kbottla, line 4445: Don't think we need submodules anymore. cardID submodule is required only for building the image which CodeBuild does.

rfisher-gds: I use it to get the `RELEASE_NAME`. Can we get that from the main repo as well?

kbottla: yes. we don't submodule for tags

jfharden:
In terms of rolling this out, we should replace the existing build-and-push-cardid-candidate with this job (please keep the old name exactly as it is, this will retain the build history).

PR #993, opened 2023-12-20
PP-11646: Change scale tasks to scale publicauth restored DB

jfharden:
We did a DR recovery of the publicauth database in perf, change the scale tasks to scale the new DB


alexbishop1 (approved):
1️⃣

PR #1007, opened 2024-01-05
Pp 12034/push multiarch stubs to deploy

jfharden:
We need to copy multi-arch images from one account to another. This adds a task (and uses it in e2e-helpers for stubs) which:

1. uses 1 role in a source account, logs in to ECR
2. pulls the images XX-candidate-<ARCH> (both amd64 and armv8)
3. uses a second role in a destination account and logs into ECR again with that
4. Pushes the images to the destination account
5. Creates a manifest in the destination account of XX-release


kbottla (approved):
LGTM!

PR #1012, opened 2024-01-09
PP-12051: Add pay-cli pipeline

jfharden:
Add a pay-cli pipeline, this works the same as the ones for pay-js-metrics and pay-js-commons to ensure we have the same npm release process as those packages


nlsteers (approved):
👍🏻

PR #1029, opened 2024-01-22
PP-12074: Copy multi-arch images from test to staging for adminusers

jfharden:
Copy multi-arch images for adminusers from test to staging.

I've made some of the config anchors so they can be reused in the million other places they are needed and they don't change at all.

You can see a successful copy:

Screenshot 2024-01-22 at 15 44 25

https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/push-adminusers-to-staging-ecr/builds/605

### DIFF
--- a/ci/pipelines/deploy-to-test.yml
+++ b/ci/pipelines/deploy-to-test.yml
@@ -1,3 +1,40 @@
+definitions:
+  - &assume_copy_from_test_ecr_role
+    task: assume-copy-from-ecr-test-role
+    file: pay-ci/ci/tasks/assume-role.yml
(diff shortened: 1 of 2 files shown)


ci/pipelines/deploy-to-test.yml
```diff
@@ -1,3 +1,40 @@
+definitions:
```
jfharden, line 1: These are anchors because we're going to reuse the same large unchanging blobs in every copy

rfisher-gds (approved):

PR #1031, opened 2024-01-23
BAU pin `dind` for concourse runners

nlsteers:
### WHAT

Stop dependabot from trying to update to a `dind` version that pact tests are not compatible with.


katstevens (approved):

katstevens:
Oh - I thought we merged this. No wonder the dependabot cropped up again

PR #1039, opened 2024-01-30
PP-12074-copy-all-multiarch-builds-to-prod

rfisher-gds:
Push all multiarch builds from staging to production. Adminusers has already been done, in #1037.

This is a big change, but to make reviewing as easy as possible, each application is in its own commit. (Which will be squashed on merge.)

Everything has been tested by running the relevant `push-XXXX-to-production-ecr` job in Concourse.


jfharden (approved):
👏 Nice!

All looks good to me, all the triggers seem correct and all the jobs ran fine

PR #1041, opened 2024-01-31
PP-12075-copy-adminusers-to-test-perf

rfisher-gds:
Copy an the adminusers image from production ECR back to test-perf.

I have modified `ci/tasks/copy-multiarch-image-to-other-account.yml` to allow a tag fragment other than `candidate`. Images are tagged "xxx-test", which I think is the right thing to do.

There is a lot of whitespace in this change, to make the top definitions syntactically correct.

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -1,200 +1,236 @@
-efinitions:
-  aws_assumed_role_creds: &aws_assumed_role_creds
-    AWS_ACCESS_KEY_ID: ((.:role.AWS_ACCESS_KEY_ID))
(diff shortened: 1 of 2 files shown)


jfharden:
For copying to perf, it's important to remember that XXX-candidate-armv8 XXX-candidate-amd64 and XXX-release (manifest) already exists, since the ECR repos are per-account and deploy-to-test already created them

Our deploy-to-perf pipelines also trigger from image tags of XXX-perf. The current pipeline does copy the image, but it's only so it can retag XXX-release as XXX-perf, so I think in this case you only need to create a new manifest XXX-perf which references XXX-release (I think the [manifest-retag task](https://github.com/alphagov/pay-ci/blob/master/ci/tasks/manifest-retag.yml) will do what you need 🤔 

Sorry for not discussing this earlier, my brain is still a bit all over the place :(

PR #1042, opened 2024-02-01
PP-12075-retag-adminusers-for-perf-test

rfisher-gds:
Retag the adminusers candidate in the test ECR for test-perf. Requires [this change to pay-infra](https://github.com/alphagov/pay-infra/pull/4728).

Not many adminusers releases require a database migration, and since we only keep a hundred images in our eu-west-1 test ECR, we cannot be certain that the perf-db candidate still exists there. But, we keep everything in our eu-central-1 test ECR, so this change copies the relevant candidate release across regions prior to retagging.

Concourse runs:
* https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/deploy-to-production/jobs/retag-adminusers-image-for-test-perf-db/builds/41
* https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/deploy-to-production/jobs/retag-adminusers-image-for-test-perf/builds/393

[ECR](https://eu-west-1.console.aws.amazon.com/ecr/repositories/private/223851549868/govukpay/adminusers?region=eu-west-1)

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -301,6 +301,13 @@ resources:
     source:
       repository: govukpay/frontend
       <<: *aws_test_config
+  - name: adminusers-candidate
+    type: registry-image
+    icon: docker
+    source:
+      repository: govukpay/adminusers
+      variant: candidate
+      <<: *aws_test_config
   - name: adminusers-ecr-registry-prod
     type: registry-image
     icon: docker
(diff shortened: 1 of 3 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -1547,20 +1570,49 @@ jobs:
...
+              file: assume-retag-role/assume-role.json
+              format: json
+            - load_var: candidate-image-tag
+              file: tags/candidate-tag
+      - task: retag-candidate-as-perf-in-ecr
+        file: pay-ci/ci/tasks/manifest-retag.yml
+        params:
+          DOCKER_LOGIN_ECR: 1
+          AWS_ACCOUNT_ID: "((pay_aws_test_account_id))"
+          SOURCE_MANIFEST: "((pay_aws_test_account_id)).dkr.ecr.eu-west-1.amazonaws.com/govukpay/egress:((.:candidate-image-tag))"
```
jfharden, line 1611: egress? Shouldn't this be adminusers?

jfharden (approved):

ci/pipelines/deploy-to-production.yml
```diff
@@ -1498,20 +1499,58 @@ jobs:
...
-        passed: [adminusers-db-migration-prod]
-        trigger: true
-      - task: parse-perf-release-tag
-        file: pay-ci/ci/tasks/parse-perf-db-release-tag.yml
-        input_mapping:
-          ecr-repo: adminusers-ecr-registry-prod
-      - put: adminusers-ecr-registry-test
+      - in_parallel:
+          steps:
+          - get: adminusers-ecr-registry-test
```
kbottla, line 1504: This could potentially lead to some confusion. If multiple PRs are merged, there could be a new image in test ECR by the time the first one is deployed in production.

For example `adminusers-ecr-registry-prod` could return `999` and due to a new release `adminusers-ecr-registry-test` could return `1000`

or I am missing something?

rfisher-gds: I agree. I keep changing my mind about the way this task ought to be done. Nothing seems quite right.

rfisher-gds: No, the release number of the image in test ECR comes from the release number of the image in prod ECR.

jfharden: Yeah you shouldn't need the test one, you can be sure that any XX-release in prod has an XX-candidate in test, so it should be enough to parse the XX-release from prod and just implicitly know it's there in test in order to create a new manifest from of XX-perf

jfharden (dismissed):
👍 Good job on persisting with this :) It feels good and robust done this way

ci/pipelines/deploy-to-production.yml
```diff
@@ -212,7 +234,7 @@ resources:
     icon: github
     source:
       uri: https://github.com/alphagov/pay-ci
-      branch: master
+      branch: PP-12075-retag-adminusers-for-perf-test
```
kbottla, line 237: Update branch name?

kbottla (approved):

PR #1046, opened 2024-02-07
PP-12075-retag-for-perf-test

rfisher-gds:
Make sure all our application images end up in eu-west-1 test ECR, tagged for perf-test. Each app is in a separate commit for simple reviewing, but I'm afraid it will still be a huge chore.

Note that when I ran the jobs to test them, a couple of the DB ones failed. This is because migrations aren't that common, and  in some cases we haven't run a multiarch build since the last one. This shouldn't be an issue: what's needed is already there, and the next ones will succeed.

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -309,12 +309,6 @@ resources:
       repository: govukpay/frontend
       variant: release
       <<: *aws_prod_config
-  - name: frontend-ecr-registry-test
-    type: registry-image
-    icon: docker
-    source:
-      repository: govukpay/frontend
-      <<: *aws_test_config
   - name: adminusers-ecr-registry-prod
     type: registry-image
     icon: docker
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -736,20 +620,59 @@ jobs:
...
+            file: pay-ci/ci/tasks/assume-role.yml
+            output_mapping:
+              assume-role: assume-retag-role
+            params:
+              AWS_ROLE_ARN: arn:aws:iam::((pay_aws_test_account_id)):role/concourse
+              AWS_ROLE_SESSION_NAME: retag-ecr-image-as-perf
+          - task: generate-docker-creds-config
+            file: pay-ci/ci/tasks/generate-docker-config-file.yml
+            params:
+              USERNAME: ((docker-username))
```
kbottla, line 658: Are these docker hub creds? Any reason we need them here and in all other places?

kbottla (approved):
LGTM!

ci/pipelines/deploy-to-production.yml
kbottla: Looks like it is to get pay-concourse-runner from docker hub https://github.com/alphagov/pay-ci/blob/215b89995f3224138b311d9aae43cb75be16fb72/ci/tasks/manifest-retag.yml#L7

PR #1049, opened 2024-02-12
PP-12123: Add jobs to build and push prometheus pushgateway resource

jfharden:
Add a job to the prometheus-pushgateway pipeline to build and push the prometheus-pushgateway-resource to dockerhub.

Later, when we aren't just spiking on this it would be good to add some integration tests between candidate and confirmed release, but while we are spiking this minimal setup will do.

Also bump the pushgateway to 1.7.0

### DIFF
--- a/ci/pipelines/prometheus-pushgateway.yml
+++ b/ci/pipelines/prometheus-pushgateway.yml
@@ -173,7 +173,7 @@ jobs:
         params:
           CONTEXT: prometheus-pushgateway-resource-release
           DOCKER_CONFIG: docker_creds
-          LABEL_release_numer: ((.:release-number))
+          LABEL_release_number: ((.:release-number))
           LABEL_release_name: ((.:release-name))
           LABEL_release_sha: ((.:release-sha))
           LABEL_build_date: ((.:date))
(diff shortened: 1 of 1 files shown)


ci/pipelines/prometheus-pushgateway.yml
```diff
@@ -108,4 +135,64 @@ jobs:
...
+          - load_var: release-sha
+            file: tags/release-sha
+          - load_var: date
+            file: tags/date
+      - task: build-prometheus-pushgateway-resource
+        privileged: true
+        params:
+          CONTEXT: prometheus-pushgateway-resource-release
+          DOCKER_CONFIG: docker_creds
+          LABEL_release_numer: ((.:release-number))
```
rfisher-gds, line 176: number

rfisher-gds (changes requested):
a typo

ci/pipelines/prometheus-pushgateway.yml
jfharden: Fixed 👍 Good spot

rfisher-gds (approved):

PR #1056, opened 2024-02-13
PP-12074-fix-task-ordering

rfisher-gds:
Fix ordering for certain tasks.

### DIFF
--- a/ci/pipelines/deploy-to-production.yml
+++ b/ci/pipelines/deploy-to-production.yml
@@ -792,7 +792,7 @@ jobs:
               skip_download: true
               format: oci
             trigger: true
-            passed: [smoke-test-selfservice-on-prod]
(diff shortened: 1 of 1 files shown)


ci/pipelines/deploy-to-production.yml
```diff
@@ -1811,7 +1811,7 @@ jobs:
               skip_download: true
               format: oci
             trigger: true
-            passed: [adminusers-pact-tag]
+            passed: [smoke-test-adminusers-on-prod]
```
jfharden, line 1814: This was correct before

The general progression is:

```
deploy -> smoke test -> pact tag -> retag perf
       -> db migration -> retag perf db
```

jfharden (changes requested):
Just the adminusers change that isn't quite right

jfharden (approved):

PR #1063, opened 2024-02-23
Pp 12216/send release metrics in staging and prod

jfharden:
1. I realised we have parse-staging-release-tag and parse-ecr-release-tag tasks, they both do the same thing, but the ecr release is more generalised and already works with the new metrics, so I changed all parse-staging-release-tag steps to parse-ecr-release-tag. I then further realised in some places we are loading a variable from it and not using it, so I also removed them.
2. Add plenty of anchors for allowing sending release metrics as we did in test with the added complication that for _some_ steps we send to a different channel (the anchor was previously suffixed with _p1 which doesn't make much sense since it's announcing a good thing, so I renamed it _annouce.
3. Add deployment metric sending for all 4 of the adminusers deploy-to-staging steps.

You can see them in grafana now: 

https://grafana.monitoring.pay-cd.deploy.payments.service.gov.uk/d/e5000e04-deef-4ac7-8ffc-b7364c241dd1/release-pipeline-releases

Screenshot 2024-02-23 at 16 23 17


rfisher-gds (approved):
Nice.

PR #1089, opened 2024-03-06
PP-12390: Update deploy pipelines for new egress and webhooks-egress repos and tagging

jfharden:
Webhooks-egress and egress are both in their own repos now and using consistent tagging of XX-release and alpha_release-XX.


kerrr (approved):
LGTM 🚀

PR #1094, opened 2024-03-11
PP-12402 Pkl module for prometheus-pushgateway pipeline

kbottla:
`fly -t pay-deploy set-pipeline -c ./prometheus-pushgateway.yml -p prometheus-pushgateway`

```diff
resources:
  resource prometheus-pushgateway-pipeline has changed:
  icon: github
  name: prometheus-pushgateway-pipeline
  source:
    branch: master
+   password: ((github-access-token))
    paths:
    - ci/pipelines/prometheus-pushgateway.yml
    uri: https://github.com/alphagov/pay-ci
+   username: alphagov-pay-ci-concourse
  type: git

pipeline name: prometheus-pushgateway
```

### DIFF
--- a/ci/pkl-pipelines/common/shared_resources.pkl
+++ b/ci/pkl-pipelines/common/shared_resources.pkl
@@ -24,13 +24,10 @@ payCiGitHubResource: PayGitHubResource = new {
   }
 }
 
-function payDockerHubResource(resource_name: String, repo: String, tag: String, checkInterval: String): Pipeline.Resource = new {
+function payDockerHubResource(resource_name: String, repo: String, tag: String): Pipeline.Resource = new {
   name = resource_name
   type = "registry-image"
   icon = "docker"
-  when (checkInterval.length > 0) {
-    check_every = checkInterval
-  }
   source = new {
     ["repository"] = repo
     ["tag"] = tag
--- a/ci/pkl-pipelines/common/shared_resources.pkl
+++ b/ci/pkl-pipelines/common/shared_resources.pkl
@@ -2,36 +2,120 @@ import "package://pkg.pkl-lang.org/github.com/alphagov/pkl-concourse-pipeline/pk
 import "package://pkg.pkl-lang.org/github.com/alphagov/pkl-concourse-pipeline/pkl-concourse-pipeline@0.0.1#/TaskConfig.pkl"
 
 class PayGitHubResource extends Pipeline.Resource {
-    type = "git"
-    icon = "github"
+  type = "git"
+  icon = "github"
 }
 
 anonymousConcourseRunnerResource = new TaskConfig.AnonymousResource {
-    type = "registry-image"
-    source = new {
-        ["repository"] = "governmentdigitalservice/pay-concourse-runner"
-        ["tag"] = "latest"
-    }
+  type = "registry-image"
+  source = new {
+    ["repository"] = "governmentdigitalservice/pay-concourse-runner"
+    ["tag"] = "latest"
+  }
 }
 
 payCiGitHubResource: PayGitHubResource = new {
-    name = "pay-ci"
-    source = new Mapping {
-        ["uri"] = "https://github.com/alphagov/pay-ci"
-        ["branch"] = "master"
-    }
+  name = "pay-ci"
+  source = new Mapping {
+    ["uri"] = "https://github.com/alphagov/pay-ci"
+    ["branch"] = "master"
+    ["username"] = "alphagov-pay-ci-concourse"
+    ["password"] = "((github-access-token))"
+  }
 }
 
-function PayDockerHubResource(resource_name: String, repo: String, tag: String): Pipeline.Resource = new {
-    name = resource_name
-    type = "registry-image"
-    icon = "docker"
-    source = new {
-        ["repository"] = repo
-        ["tag"] = tag
-        ["username"] = "((docker-username))"
-        ["password"] = "((docker-access-token))"
+function payDockerHubResource(resource_name: String, repo: String, tag: String, checkInterval: String): Pipeline.Resource = new {
+  name = resource_name
+  type = "registry-image"
+  icon = "docker"
+  when (checkInterval.length > 0) {
+    check_every = checkInterval
+  }
+  source = new {
+    ["repository"] = repo
+    ["tag"] = tag
+    ["username"] = "((docker-username))"
+    ["password"] = "((docker-access-token))"
+  }
+}
+
+function payGithubResource(resource_name: String, repo: String, tagRegex: String, path: String, repo_branch: String): Pipeline.Resource = new {
+  name = resource_name
+  type = "git"
+  icon = "github"
+  source = new {
+    ["uri"] = "https://github.com/alphagov/\(repo)"
+    when (tagRegex.length > 0) {
+      ["tag_regex"] = tagRegex
     }
+    when (repo_branch.length > 0) {
+      ["branch"] = repo_branch
+    }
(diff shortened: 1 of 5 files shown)


ci/pkl-pipelines/common/shared_resources.pkl
```diff
@@ -2,36 +2,129 @@ import "package://pkg.pkl-lang.org/github.com/alphagov/pkl-concourse-pipeline/pk
...
+  }
+  source = new {
+    ["repository"] = repo
+    ["tag"] = tag
+    ["username"] = "((docker-username))"
+    ["password"] = "((docker-access-token))"
+  }
+}
+
+function payGithubResource(resource_name: String, repo: String, tagRegex: String, path: String, repo_branch: String): Pipeline.Resource = new {
```
jfharden, line 42: I was going to say path can be a list, but actually looking at what we have it's almost always a single path. The couple that need it to be a list can amend it like:

```
(shared_resources.payGithubResource(...SNIP...)) {
  source {
    ["paths"]  {
       ...SNIP....
    }
}
```

kbottla: Updated

ci/pkl-pipelines/common/shared_resources.pkl
```diff
@@ -2,36 +2,129 @@ import "package://pkg.pkl-lang.org/github.com/alphagov/pkl-concourse-pipeline/pk
...
-    name = resource_name
-    type = "registry-image"
-    icon = "docker"
-    source = new {
-        ["repository"] = repo
-        ["tag"] = tag
-        ["username"] = "((docker-username))"
-        ["password"] = "((docker-access-token))"
+function payECRResource(resource_name: String, repo: String, aws_account_id: String,
+  check_interval: String?, tag: String?, variant: String?): Pipeline.Resource = new {
```
jfharden, line 84: I don't think these optional types actually work here, you can't optionally pass arguments, they are always a fixed number of args, that's why you have to call it with empty strings and check for it being empty.

It might be better to leave them off and amend the return value where you want to set it, e.g:
```
(payEcrResource("adminusers-ecr-registry-test", "govuk-pay/adminusers", "01234565")) {
  source { variant = "-release" }
}
```

Or alternatively have a couple of variations of this function, like one for variants, and one for tag?

```
function payECRResource(resource_name: String, repo: String, aws_account_id: String): Pipeline.Resource = new {
  name = resource_name
  type = "registry-image"
  icon = "docker"
  source = new {
    ["repository"] = repo
    ["aws_access_key_id"] = "((readonly_access_key_id))"
    ["aws_secret_access_key"] = "((readonly_secret_access_key))"
    ["aws_session_token"] = "((readonly_session_token))"
    ["aws_role_arn"] = "arn:aws:iam::((\(aws_account_id))):role/concourse"
    ["aws_ecr_registry_id"] = "((\(aws_account_id)))"
    ["aws_region"] = "eu-west-1"
  }
}

function payECRVariantResource(resource_name: String, repo: String, aws_account_id: String, variant: String) = 
(payECRResource(resource_name, repo, aws_account_id)) {
  source {
    ["variant"] = variant
  }
}
```

In this example I also eliminated the check_every, it's used hardly ever in our code, so might be better ammended in the caller on the rare occasion:
```
(payECRVariantResource(...SNIP_ARGS...)) { check_every = "never" }
```

ci/pkl-pipelines/common/shared_resources.pkl
kerrr: Is this what @jfharden was meaning? I took it to mean that defaulting to a single value was fine and that if we needed a list we could override. No?

ci/pkl-pipelines/common/shared_resources.pkl
kbottla: Makes sense. Created variant methods for frequently used variables for ECR and Github resources. and amended usages for additional attributes. We can create more variants if required later

jfharden: Awesome

ci/pkl-pipelines/pay-deploy/prometheus-pushgateway.pkl
```diff
@@ -0,0 +1,157 @@
+amends "package://pkg.pkl-lang.org/github.com/alphagov/pkl-concourse-pipeline/pkl-concourse-pipeline@0.0.1#/Pipeline.pkl"
+
+import "../common/shared_resources.pkl"
```
jfharden, line 3: Style guide says to have relative import paths on their on after the others https://pkl-lang.org/main/current/style-guide/index.html#imports

jfharden (commented):
An additional thing:

For the pipeline self-updating tasks can you use the shared stuff in https://github.com/alphagov/pay-ci/blob/master/ci/pkl-pipelines/common/pipeline_self_update.pkl

The reason being:

The existing self-update wont work anymore since they need to have the yaml generated from the pkl which doesn't happen in the existing pipeline.

It will also change the name of the self-update stuff but I think that's ok, and actually nice to have them the same in every pipeline anyway.

It should be as simple as:

```

import "../common/pipeline_self_update.yml"

resources = new {
  ..SNIP..
  pipeline_self_update.PayPipelineSelfUpdateResource("pay-deploy/prometheus-pushgateway.pkl", "master")
}

jobs = new {
  ..SNIP..
  pipeline_self_update.PayPipelineSelfUpdateJob("pay-deploy/prometheus-pushgateway.pkl"
}
```

kbottla:
@jfharden updated to use self-update pipeline. changes will

```diff
fly -t pay-deploy set-pipeline -c ./prometheus-pushgateway.yml -p prometheus-pushgateway --dry-run
resources:
  resource prometheus-pushgateway-pipeline has been removed:
- icon: github
- name: prometheus-pushgateway-pipeline
- source:
-   branch: master
-   paths:
-   - ci/pipelines/prometheus-pushgateway.yml
-   uri: https://github.com/alphagov/pay-ci
- type: git

  resource pipeline-source has been added:
+ icon: github
+ name: pipeline-source
+ source:
+   branch: master
+   paths:
+   - ci/pkl-pipelines/pay-deploy/prometheus-pushgateway.pkl
+   - ci/pkl-pipelines/common/**
+   uri: https://github.com/alphagov/pay-ci
+ type: git

jobs:
  job update-prometheus-pushgateway-pipeline has been removed:
- name: update-prometheus-pushgateway-pipeline
- plan:
- - get: prometheus-pushgateway-pipeline
-   trigger: true
- - file: prometheus-pushgateway-pipeline/ci/pipelines/prometheus-pushgateway.yml
-   set_pipeline: prometheus-pushgateway

  job update-pipeline has been added:
+ name: update-pipeline
+ plan:
+ - get: pipeline-source
+   trigger: true
+ - config:
+     image_resource:
+       name: ""
+       source:
+         repository: governmentdigitalservice/pay-concourse-runner
+         tag: latest
+       type: registry-image
+     inputs:
+     - name: pipeline-source
+     outputs:
+     - name: pipeline-source
+     platform: linux
+     run:
+       args:
+       - -c
+       - pkl eval 'pay-deploy/prometheus-pushgateway.pkl' > 'pay-deploy/prometheus-pushgateway.yaml'
+       dir: pipeline-source/ci/pkl-pipelines/
+       path: ash
+   task: generate-yaml-from-pkl
+ - file: pipeline-source/ci/pkl-pipelines/pay-deploy/prometheus-pushgateway.yaml
+   set_pipeline: prometheus-pushgateway-test

pipeline name: prometheus-pushgateway

Dry-run mode was set, exiting.
```

jfharden (approved):

PR #1141, opened 2024-04-02
PP-12484-extend-test-role-duration

rfisher-gds:
Request a three hour assume-role lease. I think this will work: I didn't find any cap on duration in the `pay-cd-pay-dev-codebuild-executor-perf-tests-test-perf-1` role.

### DIFF
--- a/ci/scripts/assume-role.js
+++ b/ci/scripts/assume-role.js
@@ -8,7 +8,8 @@ const sts = new AWS.STS()
 const run = async function run () {
   const assumeRoleResponse = await sts.assumeRole({
     RoleArn: process.env.AWS_ROLE_ARN,
-    RoleSessionName: process.env.AWS_ROLE_SESSION_NAME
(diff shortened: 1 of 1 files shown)


ci/scripts/assume-role.js
```diff
@@ -8,7 +8,8 @@ const sts = new AWS.STS()
 const run = async function run () {
   const assumeRoleResponse = await sts.assumeRole({
     RoleArn: process.env.AWS_ROLE_ARN,
-    RoleSessionName: process.env.AWS_ROLE_SESSION_NAME
+    RoleSessionName: process.env.AWS_ROLE_SESSION_NAME,
+    DurationSeconds: 10800, // 3 hours
```
jfharden, line 12: We should take this as an env var and in the assume-role task take the duration as a param so we can override it just for our perf test.

I want us to actually merge these changes permanently since it's always going to be an issue for longer perf tests).

jfharden (approved):

PR #1159, opened 2024-04-12
PP-12537: Concourse pipeline to build pay-java-lambdas JARs

oswaldquek:
Pipeline to test and build pay-java-lambdas JARs. Only the relevant submodules for the last pay-java-lambda's commit are built.

Pushing the JARs to the s3 bucket(s) is another task in itself and will be in the next PR.

See example output at https://pay-cd.deploy.payments.service.gov.uk/teams/pay-deploy/pipelines/build-and-push-java-lambdas/jobs/build/builds/32:
```
FINAL BUILD OUTPUT:
ls -alh build/bin-ranges-diff
total 25M    
drwxr-xr-x    2 root     root        4.0K Apr 15 13:56 .
drwxr-xr-x    5 root     root        4.0K Apr 15 13:56 ..
-rw-r--r--    1 root     root       24.8M Apr 15 13:56 bin-ranges-diff-1.0.20240415135459.jar
ls -alh build/bin-ranges-promotion
total 25M    
drwxr-xr-x    2 root     root        4.0K Apr 15 13:56 .
drwxr-xr-x    5 root     root        4.0K Apr 15 13:56 ..
-rw-r--r--    1 root     root       24.8M Apr 15 13:56 bin-ranges-promotion-1.0.20240415135459.jar
ls -alh build/bin-ranges-transfer
total 37M    
drwxr-xr-x    2 root     root        4.0K Apr 15 13:56 .
drwxr-xr-x    5 root     root        4.0K Apr 15 13:56 ..
-rw-r--r--    1 root     root       37.4M Apr 15 13:56 bin-ranges-transfer-1.0.20240415135459.jar
```
which shows the relevant JARs have been copied to the `build` output.

### DIFF
--- a/ci/scripts/build-java-lambdas.sh
+++ b/ci/scripts/build-java-lambdas.sh
@@ -1,0 +1,41 @@
+#!/bin/ash
+# shellcheck shell=dash
+set -o pipefail
+
+cd src || exit
+
+git fetch origin main:main
+git diff --name-only --diff-filter=d main~ main | xargs -n 1 dirname | grep '^bin-ranges-' | cut -f 1 -d "/" |  uniq > submodules.txt
+echo "Submodules to build:"
+cat submodules.txt
+
+export MAVEN_HOME=/usr/lib/mvn
+export PATH=$MAVEN_HOME/bin:$PATH
+export MAVEN_REPO="$PWD/.m2"
+
+cat <<'EOF' >settings.xml
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
+https://maven.apache.org/xsd/settings-1.0.0.xsd">
+<localRepository>${env.MAVEN_REPO}</localRepository>
+</settings>
+EOF
+
+PACKAGE_VERSION="1.0.$(date +"%Y%m%d%H%M%S")"
+echo "PACKAGE_VERSION=$PACKAGE_VERSION"
+mvn versions:set -DnewVersion="$PACKAGE_VERSION"
+
+while read -r line; do
+  mvn --global-settings settings.xml clean install --projects "$line" --also-make
+done < submodules.txt
+
+while read -r line; do
+  ls -alh "$line"/target
+  mkdir ../build/"$line"
+  cp "$line"/target/"$line"-*.jar ../build/"$line"
+done < submodules.txt
+
+cd ..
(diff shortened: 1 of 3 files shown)


ci/tasks/build-java-lambdas.yml
```diff
@@ -0,0 +1,56 @@
...
+  - name: build
+params:
+  app_name:
+platform: linux
+run:
+  path: ash
+  args:
+  - -ec
+  - |
+    #!/bin/ash
```
kbottla, line 20: Can this be moved to a .sh file so it is to shellcheck and better IDE support?

oswaldquek: Is there an existing pattern to do this? All the task files in the `ci/tasks` directory have the ash/bash script in the yml file.

kbottla: https://github.com/alphagov/pay-ci/blob/a233afcad90d2c8f1d89ea19ac939091012ef1c8/ci/pipelines/deploy-to-staging.yml#L3662

^ is node though. 

bash could follow similar pattern. One in pkl https://github.com/alphagov/pay-ci/blob/f06052e1757da0ecfd365006753690728e8da998/ci/pkl-pipelines/common/shared_resources_for_deploy_pipelines.pkl#L119

jfharden: There's an example in this also open PR: https://github.com/alphagov/pay-ci/pull/1163/files#diff-21cd77e1d1e1ab24276cc0488b99256c3499af730656130b6f8c902293c35ac8R16

ci/scripts/build-java-lambdas.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/ash
+# shellcheck shell=dash
+set -euo pipefail
+
+cd src || exit
```
jfharden, line 5: This should be `exit 1` so it exits with an error. Otherwise if the src directory doesn't exist the job will still complete because this exit exit's with success. But actually because you have `set -e` you can get rid of the || exit ecompletely since if the cd fails the script will exit as an error:

```suggestion
cd src
```

jfharden (approved):
Approved knowing there are follow-on PRs coming.

The pipeline needs a self-update task in it. As an aside if you convert it to pkl you could re-use the self-update stuff :)

oswaldquek:
> Approved knowing there are follow-on PRs coming.
> 
> The pipeline needs a self-update task in it. As an aside if you convert it to pkl you could re-use the self-update stuff :)

I'm going on hols on Thurs and I probably won't have time to learn pkl... is an example I can copy pasta?

PR #1171, opened 2024-04-18
PP-12248: Check for ARM and AMD images in docker manifest check

jfharden:
Update manifest check to ensure manifests have both ARM64 and AMD64 images in them.

We saw a failure after https://github.com/alphagov/pay-nginx-forward-proxy/pull/92 was merged since although the FROM line was a sha pin to a manifest, the manifest did not have an ARM64 image in, so after merge the build failed.

This check should prevent that in future.

### DIFF
--- a/.github/workflows/_validate_docker_image_is_manifest.yml
+++ b/.github/workflows/_validate_docker_image_is_manifest.yml
@@ -50,6 +50,7 @@ jobs:
             DIGEST=$(cut -f 2 -d "@" <<<"$BASE_CONTAINER_DEFINITION")
 
             if [ -z "$DIGEST" ]; then
+              echo
               echo "Error There was no image digest in '$BASE_CONTAINER_DEFINITION'"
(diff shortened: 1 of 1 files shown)


rfisher-gds (changes requested):
It would be polite to clean up the TMPFILE

jfharden:
> It would be polite to clean up the TMPFILE

@rfisher-gds I don't think it needs to be since it's run in an ephemeral runner, however it doesn't hurt and it is polite :) So i've amended it to remove the TMPFILE on exit

jfharden:
I've pushed another commit to https://github.com/alphagov/pay-nginx-forward-proxy/pull/99 to make sure the new changes work ok still

jfharden:
> I've pushed another commit to [alphagov/pay-nginx-forward-proxy#99](https://github.com/alphagov/pay-nginx-forward-proxy/pull/99) to make sure the new changes work ok still

It worked :)

rfisher-gds (approved):

PR #1186, opened 2024-05-02
PP-12584 Reusable workflows for Java tests

kbottla:
## WHAT
- Added reusable GitHub actions workflow to
   - run integration tests for Java apps and publish pact (based on input param)
   - run app as provider contract tests
   - run pact provider tests for consumer

Used workflows in adminusers PR
- https://github.com/alphagov/pay-adminusers/pull/2486
- Test run: https://github.com/alphagov/pay-adminusers/actions/runs/8923066283 
image

- and a PR demonstrating failed pact tests on PR https://github.com/alphagov/pay-adminusers/pull/2491

### DIFF
--- a/.github/workflows/_run-app-as-provider-contract-tests.yml
+++ b/.github/workflows/_run-app-as-provider-contract-tests.yml
@@ -1,0 +1,68 @@
+# This workflow is meant to be run on a pact provider build. It will check pacts
+# all the consumers that depend on it against itself (the provider).
+#
+name: Provider Contract Tests
+
+on:
+  workflow_call:
+    inputs:
+      java_version:
+        type: string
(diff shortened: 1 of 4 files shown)


.github/workflows/_run-app-as-provider-contract-tests.yml
```diff
@@ -0,0 +1,68 @@
+# This workflow is meant to be run on a pact provider build. It will check pacts
```
kbottla, line 1: This is same as `_run-provider-contract-tests.yml` but with a different name. `_run-provider-contract-tests.yml` will be removed when it is no longer used.

jfharden (approved):
👏

PR #1254, opened 2024-07-03
PP-12796 Job to set support (rota) topics

kbottla:
## WHAT

- Job and related scripts to update support/incident channel topics for rota changes

### DIFF
--- a/ci/tasks/support-rota-notifications.yml
+++ b/ci/tasks/support-rota-notifications.yml
@@ -1,0 +1,38 @@
+---
+platform: linux
+
+image_resource:
+  type: registry-image
+  source:
+    repository: node
+    tag: "18"
+
+inputs:
+  - name: pay-ci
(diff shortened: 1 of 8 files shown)


ci/tasks/support-rota-notifications.yml
```diff
@@ -0,0 +1,37 @@
...
+  PD_OOH_COMMS_SCHEDULE_ID:
+  OPERATION:
+  ACCESS_CONTROL_USER_CONFIG_FILE_PATH:
+
+run:
+  path: /bin/sh
+  args:
+    - -c
+    - |
+      cd pay-ci/ci/scripts/support-rota-notification
```
jfharden, line 34: You should add a set -eu here, otherwise even if the npm install fails it will continue and try and execute the index script

jfharden (approved):

PR #1309, opened 2024-09-05
Pp 13158/bump docker scout cli version

jfharden:
Bump the version of docker scout cli
When I did a test build docker also gave me a warning about using CMD of the non-json format:

```
 1 warning found (use docker --debug to expand):
 - JSONArgsRecommended: JSON arguments recommended for CMD to prevent unintended behavior related to OS signals (line 59)
```

In our case it doesn't matter, but I'd rather not have a noisy warning all the time, it will make it harder to heed warnings that do matter, so change it anyway.


SandorArpa (approved):
LOKTM

PR #1329, opened 2024-10-31
PP-12686: Shared workflow for dependency reviews on PRs

katstevens:
Introduce [actions/dependency-review](https://github.com/actions/dependency-review-action) as a shared workflow, based on [GOV.UK's equivalent shared workflow](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/dependency-review.yml).

This is intended to replace Snyk PR checks which detect if a vulnerable dependency is about to be introduced via a pull request. This is currently only available for our public repositories. 

GOV.UK's implementation includes post-merge checks on the default branch. I've intentionally left these out, as these are covered by existing Dependabot alerts and I want to closely replicate our existing workflow.

I've added in two inputs to allow overrides for:
- minimum severity level 
- whether to 'warn only' and not block merging 

I'll test this by raising a temporary PR on another repo, that will attempt to downgrade a previously-fixed vulnerability.

### DIFF
--- a/.github/workflows/_run-dependency-review.yml
+++ b/.github/workflows/_run-dependency-review.yml
@@ -1,0 +1,33 @@
+name: Github Actions - Run Dependency Review
+
+on:
+  workflow_call:
+    inputs:
+      minimum_severity:
+        type: string
+        required: false
+        default: 'low'
+        description: Override default severity rating (options are 'low', 'moderate', 'high', 'critical')
+      warn_only:
+        type: boolean
+        required: false
+        default: false
+        description: Set to true to allow the action to complete with success status, regardless of findings
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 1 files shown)


katstevens:
Tested by creating a branch that temporarily downgraded `body-parser` on [pay-products-ui](https://github.com/alphagov/pay-products-ui/actions/runs/11614019923/job/32341198511?pr=2722). The workflow successfully detected the vulnerability and failed the check:

Screenshot 2024-10-31 at 14 39 19

Interestingly, when trying with a different dependency downgrade, the OpenSSF Scorecard was too low and did not meet the failure threshold ([see details](https://github.com/alphagov/pay-products-ui/actions/runs/11613876880/job/32340708527)) for 'low' severity status , and the check passed.

two-first-names (approved):

PR #1345, opened 2025-01-14
BAU: Don't validate images that are layers in the same Dockerfile

jfharden:
Only check images that are remote images, not that are references to other layers in the same manifest

### DIFF
--- a/.github/workflows/_validate_docker_image_is_manifest.yml
+++ b/.github/workflows/_validate_docker_image_is_manifest.yml
@@ -38,7 +38,13 @@ jobs:
           DOCKERFILE=${{ inputs.dockerfile }}
 
           while read -r BASE_CONTAINER_DEFINITION; do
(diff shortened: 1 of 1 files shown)


.github/workflows/_validate_docker_image_is_manifest.yml
```diff
@@ -38,7 +38,13 @@ jobs:
           DOCKERFILE=${{ inputs.dockerfile }}
 
           while read -r BASE_CONTAINER_DEFINITION; do
-            echo "Checking base container definition: $BASE_CONTAINER_DEFINITION"
+            # Only check images which refer to a remotre repository, not another layer in the same manifest
```
nlsteers, line 41:
```suggestion
            # Only check images which refer to a remote repository, not another layer in the same manifest
```
small typo

nlsteers (approved):
type in comment, can still be merged

nlsteers (approved):

PR #1362, opened 2025-02-28
Revert "PP-13604 Upgrade docker-dind base image for concourse-runner"

katstevens:
Reverts alphagov/pay-ci#1361

Unfortunately it looks like this still a breaking change.

While the e2e tests pass, it looks like images built with this version of concourse-runner are missing a layer and cannot be pushed to ECR. [See an example here](https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/pr-ci/jobs/ledger-e2e/builds/2467).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,7 @@ updates:
     time: "03:00"
   ignore:
     - dependency-name: "docker"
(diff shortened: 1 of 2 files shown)


SandorArpa (approved):
LGTM

PR #1403, opened 2025-06-03
[WIP] PP-14175: Fix local e2e container healthchecks

katstevens:
The containers that use the reverse_proxy (frontend and selfservice) are failing the healthchecks when run locally, as the healthcheck is looking for the direct port instead of the proxy.

Not sure if this config will work on Concourse so pushing this branch to it a go.


katstevens:
Going to close this for now - can't recreate the bug outside my local machine.

PR #1443, opened 2025-09-22
[BAU] Clean up github-export-sbom.sh

two-first-names:
The API call from GitHub wraps the SBOM, unwrap it, as well as cleaning up some of the shell script.


manishlad-cs (approved):
Happy with this change.

PR #1502, opened 2026-04-16
PP-15163 Add contents:read permission to CodeQL workflow

kbottla:
## WHAT
- Adds `content:read` permissions to the workflow, so the CodeQL job can run in private repositories


AbdirizakIdris (approved):
Looks good to me

PR #1508, opened 2026-04-24
PP-15194 Use endtoend tag in zap docker-compose

kbottla:
## WHAT
- There is no `tag_endtoendzap` tag being set by end-to-end test pipelines, and also the CodeBuild project doesn't define an equivalent env variable https://github.com/alphagov/pay-infra/blob/9215ab5ae62364ae443d2400d2ede6ff2d1b6b03/provisioning/terraform/modules/pay_codebuild/endtoend_tests/buildspec.yml#L5

### DIFF
--- a/ci/tasks/endtoend/zap/docker-compose.yml
+++ b/ci/tasks/endtoend/zap/docker-compose.yml
@@ -314,7 +314,7 @@ services:
       driver: "json-file"
 
   endtoend:
-    image: ${DOCKER_REGISTRY_URI:-}${repo_endtoendzap:-govukpay/endtoend-zap}:${tag_endtoendzap:-latest}
+    image: ${DOCKER_REGISTRY_URI:-}${repo_endtoendzap:-govukpay/endtoend-zap}:${tag_endtoend:-latest}
     env_file: ../docker-config/endtoend.env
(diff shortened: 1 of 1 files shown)


benjamb (approved):

ci/tasks/endtoend/zap/docker-compose.yml
```diff
@@ -314,7 +314,7 @@ services:
       driver: "json-file"
 
   endtoend:
-    image: ${DOCKER_REGISTRY_URI:-}${repo_endtoendzap:-govukpay/endtoend-zap}:${tag_endtoendzap:-latest}
+    image: ${DOCKER_REGISTRY_URI:-}${repo_endtoendzap:-govukpay/endtoend-zap}:${tag_endtoend:-latest}
```
benjamb, line 317: Cool! I'd just spotted this, but wasn't sure if we wanted to ensure `tag_endtoendzap` was actually set, given they are in separate projects/repos, or whether to just do this.

kbottla: There will be several places that need to be updated and need to be done in order. We can look into it separately.

PR #1533, opened 2026-06-11
Add github-user-management Concourse pipeline (PP-15179).

mabdullah997861:


ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
+resources:
+  - name: configuration-repo
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/govuk-pay/configuration.git
+      branch: feature/PP-15179-app-auth
```
benjamb: I don't think this branch is necessary.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+      username: "alphagov-pay-ci-concourse"
```
benjamb: This feels like something that should be behind a variable, not sure if we have a variable set for this somewhere already?

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+      password: ((github-access-token))
+  - name: pay-access-control
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/alphagov/pay-access-control
+      branch: main
+      username: "alphagov-pay-ci-concourse"
```
benjamb: As above.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+        file: assume-role/assume-role.json
+        format: json
+      - task: terraform-plan
+        config:
+          platform: linux
+          image_resource:
+            type: registry-image
+            source:
+              repository: hashicorp/terraform
+              tag: latest
```
benjamb: We want this to use the version of terraform configured within the repository. We should add a `.terraform-version` file within the `configuration` repo à la `pay-infra`.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+            AWS_SESSION_TOKEN: ((.:role.AWS_SESSION_TOKEN))
+            AWS_REGION: eu-west-1
+            GITHUB_OWNER: govuk-pay
+            TF_VAR_github_app_id: ((github_app_id))
+            TF_VAR_github_app_installation_id: ((github_app_installation_id))
+            RAW_GITHUB_APP_PEM: ((github_app_pem))
+          run:
+            path: sh
+            args:
+              - -exc
```
benjamb: We probably don't want to specify `-x` here, it's useful for debugging but a little noisy otherwise.

Would be useful to set `-u` as well.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+              - |
+                apk add --no-cache python3 py3-pip bash aws-cli || true
```
benjamb: We should probably use the `pay-concourse-runner` here, which includes some of the above dependencies.

I also don't think the `|| true` makes a lot of sense here.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+                ln -sf /usr/bin/python3 /usr/local/bin/python || true
```
benjamb: Why is this symlink required? I think we should remove this.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+            - name: configuration-repo
+            - name: pay-access-control
+            - name: pay-ci
+          params:
+            AWS_ACCESS_KEY_ID: ((.:role.AWS_ACCESS_KEY_ID))
+            AWS_SECRET_ACCESS_KEY: ((.:role.AWS_SECRET_ACCESS_KEY))
+            AWS_SESSION_TOKEN: ((.:role.AWS_SESSION_TOKEN))
+            AWS_REGION: eu-west-1
+            GITHUB_OWNER: govuk-pay
+            TF_VAR_github_app_id: ((github_app_id))
```
benjamb: According to this docs this can be `GITHUB_APP_ID`.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+            TF_VAR_github_app_installation_id: ((github_app_installation_id))
```
benjamb: According to the docs this can be `GITHUB_APP_INSTALLATION_ID`.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+            args:
+              - -exc
+              - |
+                apk add --no-cache python3 py3-pip bash aws-cli || true
+                ln -sf /usr/bin/python3 /usr/local/bin/python || true
+
+                set +x
+                printf '%b' "$RAW_GITHUB_APP_PEM" > /tmp/github-app.pem
+                chmod 600 /tmp/github-app.pem
+                export TF_VAR_github_app_pem="$(cat /tmp/github-app.pem)"
```
benjamb: According to the docs this can be `GITHUB_APP_PEM_FILE`.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+          run:
+            path: sh
+            args:
+              - -exc
+              - |
+                apk add --no-cache python3 py3-pip bash aws-cli || true
+                ln -sf /usr/bin/python3 /usr/local/bin/python || true
+
+                set +x
+                printf '%b' "$RAW_GITHUB_APP_PEM" > /tmp/github-app.pem
```
benjamb: Can we not use `$RAW_GITHUB_APP_PEM` directly?

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+                chmod 600 /tmp/github-app.pem
+                export TF_VAR_github_app_pem="$(cat /tmp/github-app.pem)"
+                set -x
+
+                aws sts get-caller-identity
```
benjamb: Good for debugging but we should probably drop this.

ci/pipelines/github-user-management.yml
```diff
@@ -0,0 +1,85 @@
...
+
+                python3 -m pip install --break-system-packages -r pay-access-control/scripts/requirements.txt
+                sed -i '/python -m pip install/d' configuration-repo/scripts/generate-csvs.sh
```
benjamb: I think we need a PR in the configuration repo to separate the dependency installation from the running of the script. Using `sed` here to remove it feels a bit hacky.

benjamb (changes requested):

benjamb:
Closing this since it was superseded by #1557.

PR #1556, opened 2026-07-07
PP-15179 Upgrade to latest version of pkl

benjamb:
This is partially blocking development of the new GitHub org pipeline, since the newer pkl formatter I'm using locally is causing the older version to fail to evaluate due to some fragile newline parsing:
```
–– Pkl Error ––
No viable alternative at input `.PayPipelineSelfUpdateResource(\n    "pay-deploy/github-user-management.pkl",\n    "PP-15179-add-github-org-pipeline",\n  )`.
```


SandorArpa (approved):
LGTM

PR #1563, opened 2026-07-17
[PP-15612] added test to verify octokit in node-runner

foxmaltas-gds:
Added a javascript test to verify that octokit can fetch a branch by tag in node-runner

### DIFF
--- a/.github/workflows/test-node-runner-octokit.yml
+++ b/.github/workflows/test-node-runner-octokit.yml
@@ -1,0 +1,28 @@
+name: Test octokit/rest in node-runner
+
+on:
+  pull_request:
+    paths:
+      - ci/docker/node-runner/**
+      - .github/workflows/test-node-runner-octokit.yml
+
+jobs:
+  test-octokit-in-node-runner:
+    runs-on: ubuntu-latest
+    name: Run test script for octokit/rest
+    
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      - name: Setup Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e
+        with:
+          node-version-file: 'ci/scripts/.nvmrc'
+          cache: 'npm'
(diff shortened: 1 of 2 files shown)


.github/workflows/test-node-runner-octokit.yml
```diff
@@ -0,0 +1,20 @@
+name: Test octokit/rest in node-runner
+
+on:
+  pull_request:
+    paths:
+      - ci/docker/node-runner/**
```
benjamb, line 6: You probably also want to include the path to this file.

.github/workflows/test-node-runner-octokit.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  test-octokit-in-node-runner:
+    runs-on: node-runner
+    name: Run test script for octokit/rest
+    env:
+      working-directory: ci/scripts
```
benjamb: Assuming this just sets environment variables, given this isn't used below, I'd drop this.

.github/workflows/test-node-runner-octokit.yml
```diff
@@ -0,0 +1,20 @@
...
+    
+    steps:
+      - name: Checkout
+        uses: actions/checkout
```
benjamb: Could you pin this to a sha? Use the same version as in the `test-semver.yml` file.

.github/workflows/test-node-runner-octokit.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Run test
```
benjamb, line 26: You probably need `actions/setup-node` before this. You can copy this from `test-semver.yml` as well.

.github/workflows/test-node-runner-octokit.yml
```diff
@@ -0,0 +1,20 @@
+name: Test octokit/rest in node-runner
+
+on:
+  pull_request:
+    paths:
+      - ci/docker/node-runner/**
+
+jobs:
+  test-octokit-in-node-runner:
+    runs-on: node-runner
```
benjamb: This refers to the github runner, not the container. This should instead be `ubuntu-latest`.

PR #1571, opened 2026-07-24
[PP-15612] Add missing test file

foxmaltas-gds:
The test file was missed from the previous PRs, so this adds it. The test checks if octokit can download a git reference by tag


benjamb (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay.

PR #1573, opened 2026-07-27
[PP-15547] Added reusable workflow for checking terraform formatting

foxmaltas-gds:
This will be used by the "pay-infra", "pay-concourse", and "pay-concourse-deployment" repos as they all use the same workflow/job for checking terraform formatting. It was modified from the `run-terraform-tests.yaml` workflow in the `pay-infra` repo [here](https://github.com/alphagov/pay-infra/blob/master/.github/workflows/run-terraform-tests.yml)

### DIFF
--- a/.github/workflows/_run-terraform-tests.yml
+++ b/.github/workflows/_run-terraform-tests.yml
@@ -1,0 +1,73 @@
+name: Terraform Tests
+
+on:
+  workflow_call:
+  
+jobs:
+  terraform-fmt:
+    name: Terraform format check
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: "0"
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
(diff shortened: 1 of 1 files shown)


two-first-names (approved):

.github/workflows/_run-terraform-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+            echo
+            exit 1
+          }
+
+          export PATH="${GITHUB_WORKSPACE}/.tfenv/bin:$PATH"
+          export CHECKPOINT_DISABLE=true # Stop TF showing "your version is out of date" warnings
+
+          trap help_with_failure ERR
+
+          git diff --name-only origin/master provisioning/**/*.tf | while read -r TF_FILE_PATH; do
```
benjamb, line 54: We either want to make this path generic, i.e. just `**/*.tf`, or accept the path as an input.

benjamb: We likely want the branch we're comparing against, in this case `origin/master`, to be converted to an input parameter as well.

foxmaltas-gds: Resolved in this PR [https://github.com/alphagov/pay-ci/pull/1574](https://github.com/alphagov/pay-ci/pull/1574)

PR #1578, opened 2026-07-29
[PP-15555] Move generate csvs dependancy to concourse runner

foxmaltas-gds:
The "generate-github-acl-csvs.sh" script only needs the python yaml package, so this change installs it in the concourse runner docker image instead of having the script install it using pip. Further changes will be done as part of PP-15495

### DIFF
--- a/ci/docker/concourse-runner/Dockerfile
+++ b/ci/docker/concourse-runner/Dockerfile
@@ -14,6 +14,7 @@ RUN apk add --no-cache \
   gcompat \
   git \
   python3 \
+  py3-yaml \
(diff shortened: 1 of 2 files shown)


ci/docker/concourse-runner/Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN apk add --no-cache \
   npm \
   maven \
   tini \
+  py3-yaml \
```
benjamb: Minor nit: could you put this just under the `python3` dependency?

benjamb (approved):

